PR #11, opened 2019-10-09
feat: add stopped state and pending/stop events

sublimator:


docs/api.md
```diff
@@ -3,61 +3,101 @@ id: api
...
+### Stopped
 
 ```ts
-document.monetization && document.monetization.state === 'pending'
+document.monetization && document.monetization.state === 'stopped'
 ```
 
-The user is capable of Web Monetization. The [Web Monetization Service Provider](/#providers) is starting to send a streaming payment with Interledger.
+The user's browser is capable of Web Monetization.
+If a monetization tag is found **AND** the [Web Monetization Service Provider](/#providers) is enabled/active/etc the `document.monetization.state` will transition to `pending` as STREAMing is initiated.
```
sublimator: Followed the prevailing "STREAMing" convention in the document

docs/api.md
```diff
@@ -3,61 +3,101 @@ id: api
...
 
-You can access Web Monetization-related events and state through a Javascript
+You can access Web Monetization related events and state through a Javascript
 API on `document.monetization`.
 
 ## Javascript API
 ```ts
 document.monetization: EventTarget
-document.monetization.state: string
+document.monetization.state: 'stopped' | 'pending' | 'started'
```
sublimator: Are we sure we don't want to just leave `pending` as is ( essentially stopped ) and add an `initializing` state (essentially pending ) 

At some point we'll have to follow up with some breaking changes to match the W3C standard, so it would just be a temporary "strangeness"

If we are concerned about breaking things in the short term, that would help minimize changes.

docs/api.md
```diff
@@ -103,25 +174,29 @@ document
 
 *Event Object*
 
-Field | Value | Details 
+Field | Value | Details
 --- | --- | ---
+paymentPointer | String | The `paymentPointer` matches the one in the `<meta name="monetization">` tag.
```
sublimator: Added these fields as per the impl and current explainer example

sublimator:
I just went with the existing structures. Feel free to push a fix commit or
I will get to it tomorrow.

We should get a CoilPreview extension going.

docs/api.md
sublimator: OK then

sublimator:
> I am still worried about the privacy implications of "pending" but let's experiment and get feedback.

@adrianhopebailie 
Could you spell the implications out ? I'm missing something.

I can't imagine a provider configured to block a site would start a stream, set the state to "pending", then stop it once the "rules" kicked in. It would start in the 'stopped' state and just stay there, as we assume a browser with WM implemented, yet no provider configured would. 

?

adrianhopebailie:
> Could you spell the implications out ? I'm missing something.

The more state changes we have the easier it is for a website to infer things about the user based on the timing and order of these changes.

That is leaking info about the user.

We should consider normative requirements around how these state changes occur so that it is hard for a website to infer, for example, the difference between a user that doesn't have monetization setup vs a user that does but has disabled it for the current site.

As I said to @sharafian I don't think it needs to block this PR but we need to explore this and at a minimum make a mention in our Security and Privacy considerations for implementors.

cc @tobie to keep in mind for the spec

sharafian:
I think the assumption is that if you integrated this into browsers then everybody with a WM capable browser who had not enabled WM would be in the "stopped" state.

The case that a website could maybe guess it was turned off would be when the user is paying and then stops in the span of one session, but that's something they can always detect server side.

sublimator:
> that's something they can always detect server side.

ja

PR #36, opened 2019-11-14
docs: add web monetization examples

sharafian:
Added some examples on how to do exclusive content with Web Monetization. Also included the examples in the static folder so users can try them out.

edit: also added a remove ads example
edit: added example counter and start/stop wm example

### DIFF
--- a/website/static/examples/show.html
+++ b/website/static/examples/show.html
@@ -1,5 +1,5 @@
 <head>
-  <meta name="monetization" content="$twitter.xrptipbot.com/Coil">
+  <meta name="monetization" content="$twitter.xrptipbot.com/Interledger">
 
   <style>
     .hidden {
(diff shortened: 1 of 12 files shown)


website/static/examples/show.html
```diff
@@ -0,0 +1,54 @@
+<head>
+  <meta name="monetization" content="$twitter.xrptipbot.com/Interledger">
```
traviscrist: Should this use the coil payment pointer, not sure this one actually works...

sharafian: https://twitter.com/Interledger it's certainly a legitimate twitter account. It feels weird to make the examples on webmonetization.org pay to Coil. If there were a verified charity I'd be good with using it in examples

sharafian: I'll also include a note that this should be changed to your own payment pointer

traviscrist: I agree that it ideally should not be coil, can we make both go to interledger then? I noticed the other one went to Coil.

PR #37, opened 2019-11-15
fix: edited web mon examples

melissahenderson:


melissahenderson:
Please merge into yours if this looks good. Also, there's the Coil payment pointer in remove_ads.html. Not sure if you wanted to change that.

docs/exclusive-content.md
```diff
@@ -6,16 +6,16 @@ sidebar_label: Exclusive Content
 
 One of the perks of Web Monetization is that its JavaScript API can be used to
 make your page respond to Web Monetization. You can reward the people who
-support your site by giving Web Monetized viewers exclusive content.
+support your site by giving Web-monetized viewers exclusive content.
```
sublimator: I know it's probably correct, but "Web-monetized" just looks off to me :)

docs/remove-ads.md
```diff
@@ -112,15 +115,11 @@ window.addEventListener('load', () => {
...
+If the visitor _does_ have Web Monetization, the three-second timer starts.
+We check whether the `monetizationstart` event has fired when the timer is up.
+If Web Monetization hasn't initialized, the visitor is shown the ads.
 
 > You might think of using `document.monetization.state` instead of remembering
-> `hasPaid`. But the state can go back to being stopped or pending if the user
-> backgrounds the tab. If they background your tab when the 3 seconds are over
-> then a legitimately Web Monetizated user might would get shown ads! So
-> keeping a `hasPaid` variable is better.
+> `hasPaid`. But, the state can go back to being `stopped` or `pending` if the user
```
sublimator: Given this, should this be tracked on the document.monetization object ?

docs/remove-ads.md
```diff
@@ -112,15 +115,11 @@ window.addEventListener('load', () => {
 })
 ```
 
-As soon as the page loads, any visitor who does not have Web Monetization
-(`!document.monetization`) sees ads immediately.
-
-If the visitor _does_ have Web Moentization we start our three second timer.
-When the timer is up we check whether the `monetizationstart` event has fired.
-If Web Monetization hasn't initialized, we show them the ads.
+If the visitor _does_ have Web Monetization, the three-second timer starts.
```
sharafian: I would put this with the previous paragraph before the code block, because both apply to the same block of code. Or we could split the code block into

```js
window.addEventListener('load', () => {
  if (!document.monetization) {
    showAds()
```
and
```js
  } else {
    setTimeout(() => {
      if (!hasPaid) {
        showAds()
      }
    }, 3000)
  }
})
```

docs/remove-ads.md
sharafian: I would say similar to "total," if it can be derived easily we don't need to include it in the WM API. This also gives more control to the page on what the scope of "hasPaid" is.

sharafian:
Other than the one change this looks good to merge. Thanks for looking this over and giving feedback!

sublimator:
@sharafian

It’s true, it’s not necessary to provide information that can be derived.
You could say the same for the paymentPointer/requestId in the
monetizationprogress event detail.

How often would the information be used/necessary though? Something like
”document.monetization.current” could be quite convenient, which isn’t a
bad thing for an API.

It’s easier to add than remove api surface though so I understand the
reluctance.

docs/exclusive-content.md
melissahenderson: After the new Coil pages are up, I'll need to edit the web mon docs. I want it to be consistent with what Fabian shows. I'm pretty sure that includes dropping the hyphen for that exact reason. It's usually correct to include, but doesn't look all that great.

docs/remove-ads.md
melissahenderson: Fixed

Issue #41, opened 2019-12-06
Explain how browsers should handle multiple or dynamic meta tag for monetization


fred-wang:
https://webmonetization.org/specification.html#flow says

> There MUST be only one <meta name="monetization"> tag at any given time

so I assume that when there are multiple tags, monetization should just be ignored. However, there are probably alternatives e.g. consider multiple monetizations or just the first one found. I guess it would be worth getting feedback from browser developers about the easiest option to implement.

Similarly, I'm not sure what happens when the meta tag is added/removed dynamically via javascript after page load... Should only a meta tag at page load considered?

Note: If both meta and link elements are supported (  #19 ), one can also specify multiple monetizations with different elements, so that would need to be clarified too. I guess some page authors would like to specify both versions for better interoperability and just ignoring everything seems wrong in that case.

sublimator:
Hi @fred-wang 

I can't speak to future plans, but currently with the proof-of-concept working implementations ( polyfill script and extension)  we:
* handle only `<meta>` tags, not `<link>`s
* track monetization meta[name=monetization] with a MutationObserver on the head and if there is more than one it will throw an Error ( read: silently do nothing except log to the console )
* support dynamic meta tags

fred-wang:
Thanks for the explanation @sublimator ; this matches how I understood the current status.

Reading https://webmonetization.org/specification.html#flow again, one can put a `<meta name="monetization" content="URL1">`, remove it, then add a new `<meta name="monetization" content="URL2">`. And this will trigger two payment requests for URL1 and URL2. Is it intended? Why not allow multiple meta tags at page load in the first place, then?

I think mechanism will have to be implemented in the UI to avoid too many noise for users in any case and allow them to discard or accept several payments in one go. From my quick look at chromium's code, each meta tag has their content parsed independently, so it's probably easier if the 'There MUST be only one meta name="monetization" tag at any given time' condition is dropped and the multiple requests handle specially (just a guess though).

Additionally, IIUC the spec allows this flow:

A. The user agent sets document.monetization.state to started (for the URL1 request)
B. The user agent sets document.monetization.state to started (for the URL2 request)
C. The user agent sets document.monetization.state back to pending (for the URL1 request)
D. The user agent sets document.monetization.state back to pending (for the URL2 request)

So during steps C and D, request for URL2 is still running but document.monetization.state is pending, which is probably not intended. I think either document.monetization should be a map from URLs to states or requests should be queued and executed one-by-one at step 3.

[...]

PR #42, opened 2019-12-12
feat: add finalized: boolean to monetizationstop event

sublimator:
### monetizationstop finalized property

Related to https://github.com/interledger/webmonetization.org/issues/39

Allows determining when a "monetization" (that which is identified by a requestId) has completely finished without resorting to [looking at the document](https://github.com/sharafian/react-web-monetization/blob/8a86b37b237c18f461bdaa5d41caa6fa841ecbd3/src/global.js#L95-L99) . 

Due to the fact that we always create a new `requestId` for any "new" meta tag we can be sure that no more events will be emitted for a given requestId when we set `"finalized": true`.  We don't track unique payment pointers per page load, so if you removed a meta tag with paymentPointer $a, then added a meta tag back with $a, then a new requestId would be created. Generally that would make sense in that the request would likely be for different content.

### Current usages of finalized: false

The main "use" of finalized: false (i.e. paused) currently is when a page is backgrounded and goes into the stopped state. When the page is foregrounded again, the stream is resumed using the same requestId.

### Specification of dynamic/multiple monetization tags
We may want to specify the relation of a requestId to meta tags more precisely.
#### Important
See #41 

### Note on finalized term
Not stuck on `finalized`. @sharafian suggested an inverted property like 'temporary' may be better but agreed `finalized` is a good starting point for discussion.


sublimator:
Well I won’t argue for it :)

sublimator:
@melissahenderson 
> I'm not sure if it's really necessary/helpful.
Wasn't sure if you meant the field or adding to examples ? 
I just added the field to the example event object, with `finalized: false`

matdehaast:
@sublimator does this mean if I get a `monetizationstop` without the `finalised` field it is essentially just a temp pause?

sublimator:
@matdehaast 
Yeah, it's essentially a monetizationpaused event when finalized: false  (it won't be omitted, rather set to false)

sharafian:
Is this added to WM projects' master branch? once it gets into an extension release we should merge this

sublimator:
@sharafian this is in the 0.0.44 release

sublimator:
once it gets into an extension release we should merge this
^

PR #44, opened 2019-12-18
Added digital wallet pages and formatting updates

melissahenderson:
Added digital wallet pages. Updated 'web monetized' wording. Includes Nic's finalized boolean update.


docs/api.md
```diff
@@ -174,20 +174,22 @@ document
...
 
-#### Event Object
+#### Event object
 
 | Field | Value | Details |
 | ------ | ----- | ------ |
 | paymentPointer | String | The payment account URL for the Web Monetization `meta` tag at the time when payment has stopped. |
-| requestId | String | The session ID/Monetization ID (UUID v4) [generated by the browser](/docs/explainer#flow) for the Web Monetization `meta` tag at the time when payment has stopped. |
+| requestId | String | The session ID/monetization ID (UUID v4) [generated by the browser](./explainer#flow) for the Web Monetization `meta` tag at the time when payment has stopped. |
+| finalized | Boolean | When `true`, the monetization tag has been removed or the `paymentPointer` changed. No more events with this `requestId` are expected.
```
sharafian: If this is pulling in the `finalized` feature then we'll have to wait until that is published through the extension before we merge this. So I think it's probably best to take out the finalized stuff in this version and let Niq rebase it on top of yours once this gets merged

docs/gatehub.md
```diff
@@ -0,0 +1,73 @@
...
+3. Click **\(+\)** in the upper-left.
+4. Enter your password.
+5. Enter a wallet name and select **Hosted** under **Wallet type**.
+6. Click **Create**. Your wallet is now created on GateHub.
+7. Click **Deposit/Receive** at the bottom of the screen.
+8. Select the **ILP** tab and copy your payment pointer to your clipboard.
+
+![createwallet](assets/gatehub-01.png)
+
+## Create your meta tag
```
sharafian: We may not want to describe this here. A lot of the time people will be sending their creators to this guide so they can get their payment pointer and enter it into a website (like cinnamon or into a wordpress plugin) rather than using the meta tag directly.

Maybe instead we can tell the user they can return to the site that needs their payment pointer and enter it, or add it to their own site by following the getting started tutorial?

docs/stronghold.md
```diff
@@ -0,0 +1,38 @@
+---
+id: stronghold
+title: Stronghold
+sidebar_label: Stronghold
+---
+
+Use Stronghold to have your earnings paid out in US dollars.
+
+> You won't see your payments immediately if the amount you've received is
```
sharafian: can we give the exact threshold here? or an approximation even

docs/api.md
melissahenderson: Removed.

docs/gatehub.md
melissahenderson: Updated the three wallet-specific pages.

docs/stronghold.md
melissahenderson: Added approximation.

PR #51, opened 2020-03-06
feat: add STREAM receipts

wilsonianb:
depends on https://github.com/interledger/rfcs/issues/568

Removes Web-Monetization-Id header


website/static/specification.html
```diff
@@ -609,6 +614,7 @@ <h2><dfn>MonetizationProgressEvent</dfn> interface</h2>
               readonly attribute DOMString amount;
               readonly attribute DOMString assetCode;
               readonly attribute DOMString assetScale;
+              readonly attribute DOMString? receipt;
```
melissahenderson: Should there be a question mark here? I don't know what the right format is :)

wilsonianb: The `?` is intentional since the receipt is optional.
That said I'm also not super familiar with this formatting.

sharafian: @adrianhopebailie would likely know this the best

docs/explainer.md
```diff
@@ -77,18 +77,21 @@ appropriate rate of payments to make to the site.
 3. The browser generates a unique session ID for this payment session.
 4. The browser fetches a unique destination address and shared secret for the
```
sharafian: maybe this should be "from the site's WM receiver or Payment Verifier Service"

website/static/specification.html
melissahenderson: There's also references to the Web-Monetization-Id header in the spec. That was removed from the other Web Mon doc. Should that come out?

docs/explainer.md
wilsonianb: Changed to "from site's payment pointer URL"
https://github.com/interledger/webmonetization.org/pull/51/commits/10a5fd13c2add0f347ba5e56a83fb87503998125

website/static/specification.html
wilsonianb: Thanks. Removed in https://github.com/interledger/webmonetization.org/pull/51/commits/10a5fd13c2add0f347ba5e56a83fb87503998125

adrianhopebailie: From my understanding of WebIDL @wilsonian has it right

wilsonianb:
The STREAM receipts rfc is now merged (interledger/rfcs#568).

docs/explainer.md
```diff
@@ -286,10 +289,11 @@ session.
         requestId,
         amount,
         assetCode,
-        assetScale
+        assetScale,
+        receipt
       } = event.detail
       if (
-        isValidPayment(paymentPointer, requestId, amount, assetCode, assetScale)
+        isValidPayment(paymentPointer, requestId, amount, assetCode, assetScale, receipt)
```
sharafian, line 296: I wonder if this example could be a little confusing by bringing in all these fields; technically payment pointer + receipt should be enough to verify payment

docs/api.md
```diff
@@ -217,9 +217,10 @@ document
...
-|amount | String | The destination amount received as specified in the Interledger protocol (ILP) packet. |
-|assetCode | String | The code (typically three characters) identifying the amount's unit. A unit, for example, could be a currency (USD, XRP). |
-|assetScale | Number | The number of places past the decimal for the amount. For example, if you have USD with an asset scale of two, then the minimum divisible unit is cents. |
+| amount | String | The destination amount received as specified in the Interledger protocol (ILP) packet. |
+| assetCode | String | The code (typically three characters) identifying the amount's unit. A unit, for example, could be a currency (USD, XRP). |
+| assetScale | Number | The number of places past the decimal for the amount. For example, if you have USD with an asset scale of two, then the minimum divisible unit is cents. |
+| receipt | String | base64-encoded STREAM receipt issued by the Web Monetization receiver to the Web Monetization provider as proof of the total amount received in the stream. |
```
sharafian, line 223: can we link this to the STREAM receipts spec?

wilsonianb: Added link (pending https://github.com/interledger/rfcs/pull/572)
https://github.com/interledger/webmonetization.org/pull/51/commits/9e265950615c8a6d88ba107115d960727a87cf2a

docs/explainer.md
wilsonianb: I think I kept the `amount` since it's subsequently used in the call to `suspendAdvertising`. Presumably `isValidPayment` verifies both the receipt and that `amount` matches the receipt amount.
Alternatively, receipt could just be passed to `suspendAdvertising`, or `isValidPayment` could return an amount (from the receipt).

docs/api.md
sublimator: > the total amount received in the stream

Interesting to note that this could be quite different to the total amount received for a given requestId (which is currently mapped 1:1 with a given meta tag)

docs/api.md
```diff
@@ -222,6 +222,10 @@ document
 | assetScale | Number | The number of places past the decimal for the amount. For example, if you have USD with an asset scale of two, then the minimum divisible unit is cents. |
 | receipt | String | base64-encoded STREAM receipt issued by the Web Monetization receiver to the Web Monetization provider as proof of the total amount received in the stream. |
 
+> There may be multiple STREAM connections for a single `requestId`.
```
sublimator, line 225: ooooh, this was actually noted, nice :)

docs/api.md
adrianhopebailie: True, but could the client aggregate these? That seems like a better experience for the website developer

docs/glossary.md
```diff
@@ -58,3 +58,9 @@ more info.
 Each page load or refresh is a unique session. The browser generates a
 [unique session ID](./explainer#flow) which the website can use to correlate
 incoming payments to the Web Monetization receiver to a client-side context.
+
+# Receipt Verifier
+* * *
+A service operated by the monetized website or third party that verifies
+payment receipts issued by the Web Monetization receiver and submitted to the
```
adrianhopebailie, line 65:
```suggestion
payment receipts issued by the Web Monetization receiver and emitted to the
```

docs/api.md
sublimator: @adrianhopebailie 
Could you elaborate/illustrate ?

adrianhopebailie: It would be nice if the browser could collect the receipts for a single requestId and make these available to the developer somehow to make it easier to sum them and get the real total for the requestId

wilsonianb: Receipt amounts should only be summed by the verifier, which would never be the browser.

Maybe this `receipts` field could be a list or map of the latest receipt from each STREAM connection or stream for this requestId, which could help inform a website that for some reason isn't submitting every receipts to the verifier. (If the site is only verifying every 10 receipts, they risk completely missing receipts from different connections or streams).
But I think I prefer only including the receipt corresponding the packet amount of the particular `monetizationprogress` event, in which case website should be verifying every receipt.

wilsonianb: >Maybe this receipts field could be a list or map of the latest receipt from each STREAM connection or stream for this requestId

That kind of grouping could be done with [`react-web-monetization`](https://github.com/sharafian/react-web-monetization) (by adding some receipt parsing) for the site.

wilsonianb: There might actually be cases (such as larger, discrete payments for mobile) in which we'd want to support multiple receipts in a single `monetizationprogress` (assuming `monetizationprogress` is used as opposed to a new event).

adrianhopebailie:
@sharafian and @wilsonianb are we gonna wait for this to be implemented by all receivers before we publish this?

wilsonianb:
I'm going to look into supporting multiple receipts in a single event for cases like mobile web monetization
https://github.com/WICG/webmonetization/pull/51#discussion_r445856623

wilsonianb:
After some consideration, these changes seem fine as is.
Instead of supporting multiple receipts in a `monetizationprogress` event, multiple events can just be emitted (one per stream) with the final receipt for that stream for cases like mobile web monetization where larger batched payments, which could involve multiple streams, may occur.
That use case might require changing the definition of the [event](https://webmonetization.org/docs/api#monetizationprogress)'s `amount` field, but that can happen in a future pr when it's needed.

@adrianhopebailie Let me know if these commits should be squashed and/or rebased

PR #53, opened 2020-03-12
iframes

melissahenderson:


docs/api.md
```diff
@@ -254,3 +254,16 @@ the `requestId` in the `monetizationstart` browser event.
...
+## Iframes
+You can monetize an iframe by adding `monetization` to the iframe's `allow` attribute. For example:
+
+```html
+  <iframe src = "/wm/example.htm" width = "600" height = "800" allow="monetization">
+```
+
+The iframe **must** have the meta tag in its head. In the example above, `example.htm` contains the Web Monetization `meta` tag.
+
+> Monetization is only supported on iframes in the top window, not iframes within iframes.
```
sublimator: We have nested iframe support in queue:
https://github.com/coilhq/web-monetization-projects/pull/460

docs/api.md
```diff
@@ -254,3 +254,16 @@ the `requestId` in the `monetizationstart` browser event.
...
 ```
+
+## Iframes
+You can monetize an iframe by adding `monetization` to the iframe's `allow` attribute. For example:
+
+```html
+  <iframe src = "/wm/example.htm" width = "600" height = "800" allow="monetization">
+```
+
+The iframe **must** have the meta tag in its head. In the example above, `example.htm` contains the Web Monetization `meta` tag.
```
sharafian: I would clarify that it must have the meta tag in order to monetize. No error will occur just from embedding an iframe with allow='monetization' and not having it be monetized

melissahenderson: Updated. Can you recheck please?

matdehaast:
@melissahenderson @sharafian we should update the wording to explain how iframe monetization is expected to work if the parent webpage has a monetization tag and also if there are multiple iframes.

PR #92, opened 2020-06-25
Added Brave as a Monetisation browser

justincavery:
Brave has also implemented micro payments through Uphold for YouTube and Twitch creators.


marcoscaceres:

adrianhopebailie:
@justincavery do you know if they've implemented the Web Monetization API? As far as I know they haven't.

PR #106, opened 2020-08-22
Fix typos in api docs page

dacioromero:


adrianhopebailie:

PR #109, opened 2020-09-16
feat: add probabilistic revshare tool

sharafian:
Adds a tool to create a probabilistic revshare pointer. The SPSP endpoint itself is controlled with Cloudflare workers, and the frontend is implemented via a custom page (possible thanks to docusaurus 2).

The worker is already published to webmonetization.org so the meta tags produced by this tool will work even in dev mode!

### Edit view

Screen Shot 2020-09-25 at 1 53 56 PM

### Import view

Screen Shot 2020-09-25 at 1 54 08 PM

## TODOs

- [x] manage overflowing of chart labels
- [x] import mode that takes your meta tag and fills in the inputs so you can update it
  - [x] save the revshare in localstorage?
  - [x] keep the names around so they aren't lost on import


melissahenderson:
I love it! I have one question/suggestion. Do you think we need to add info about clearing your cache to the "Your revshare data will also be saved in your browser" bit? I think the audience will likely understand, but just in case..

melissahenderson (approved):

sharafian:
True, that's possible people could get confused on losing that data if they clear their browser. But with the import tool hopefully that mitigates the issue somewhat. Let's see if it comes up

fruehle:
Copy suggestions.

**Meta Tag Generator (with Probabilistic Revshare)**
This tool allows you to create a Web Monetization meta tag which splits money between multiple Payment Pointers (Learn more).
If you make changes to an existing meta tag with this tool, make sure to replace the old meta tags with the update version.

**Editing Revshare**
Add all Payment Pointers that will receive a split of the revenue and label them (optional). By setting weights you determine what share each party will receive. You can also tweak the percentage to adjust shares.

If you want to edit and existing meta tag created with this tool click 'Import meta tag".
Note: Your inputs will be saved in your browser locally.

-----

**Import Probabilistic Revshare Meta Tag**
Copy your Web Monetization meta tag (e.g. found in your website's <head>) into the box below and click "Import" to load it into the editor.

Once you're finished editing, you must replace all old meta tags on your site with the newly generated one in order for the changes to take effect!

PR #134, opened 2020-10-23
feat: add exclusive content worker

sabinebertram:
Cloudflare worker code to generate exclusive content by encryption. Decryption only happens if payments can be verified.


workers/exclusive-content/LICENSE_MIT
```diff
@@ -0,0 +1,25 @@
+Copyright (c) 2020 Sabine Schaller <sabine@coil.com>
```
sharafian, line 1: We usually put stuff under the Apache-2.0 license. Might be a requirement of the WICG but in any case probably best to use the same license we use elsewhere in this project

workers/exclusive-content/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+This cloudflare worker will encrypt any plain text given a payment pointer. Furthermore, it will decrypt the resulting cypher text if it can verify payment by a receipt verifier.
+
+## Testing
+
+Configure your `wrangler.toml` file.
+
+```sh
+npm install
+wrangler secret put MASTERKEY
```
sharafian, line 11: do you know what the scope of this variable is? is it per-worker or per-zone?

workers/exclusive-content/README.md
```diff
@@ -0,0 +1,92 @@
...
+### `POST /encrypt`
+
+Encrypts plaintext
+
+#### Request Body
+
+JSON object containing
+| Name | Type | Description | Example |
+| -- | -- | -- | -- |
+| pp | String | (proxy) payment pointer | "\$spsp.example.com/alice" |
```
sharafian, line 38: we can get away with being more verbose than this; `plainText` and `paymentPointer` should be fine and will make life a little easier for us and any other consumers of this API

workers/exclusive-content/README.md
```diff
@@ -0,0 +1,92 @@
...
+Configure your `wrangler.toml` file.
+
+```sh
+npm install
+wrangler secret put MASTERKEY
+# enter a key to be stored with cloudflare worker
+wrangler publish
+```
+
+## Endpoints
```
sharafian, line 27: I like this detailed write-up of the API!

workers/exclusive-content/README.md
```diff
@@ -0,0 +1,92 @@
...
+| ------------ | ---------------- |
+| Content-Type | application/json |
+
+#### Response Body
+
+JSON object containing
+| Name | Type | Description | Example |
+| -- | -- | -- | -- |
+| ct | String | cypher text | "VsPD81db5Ts9rUnvxqGD5hxCRoDU+ygE44w42AoEbaKQbhGXkrpxoYwCrhrtR2MmFfOg" |
+| iv | String | initialization vector | "1603118464654" |
```
sharafian, line 53: what encoding is this? should be noted for all fields that represent binary data

workers/exclusive-content/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+#### Request Body
+
+JSON object containing
+| Name | Type | Description | Example |
+| -- | -- | -- | -- |
+| pp | String | (proxy) payment pointer | "\$spsp.example.com/alice" |
+| ct | String | cypher text | "VsPD81db5Ts9rUnvxqGD5hxCRoDU+ygE44w42AoEbaKQbhGXkrpxoYwCrhrtR2MmFfOg" |
+| iv | String | initialization vector | "1603118464654" |
+| vr | String | receipt verifier endpoint | "https://verifier.example.com" |
```
sharafian, line 72: I can pass in any endpoint I like here and have it say that I paid. Should probably be an env var to the worker instead. Another option would be to specify the verifier at encryption time and include it in the ciphertext (you could encrypt a json object to accomodate several fields) so that it can't be tampered with

workers/exclusive-content/README.md
```diff
@@ -0,0 +1,92 @@
...
+| bi | String | balance id used to credit the receipts | "123" |
```
sharafian, line 73: if we're dropping balance logic from the verifier we might want to just replace this with the receipt itself. Initial version could probably just unlock for any amount

workers/exclusive-content/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+Encrypts plaintext
+
+#### Request Body
+
+JSON object containing
+| Name | Type | Description | Example |
+| -- | -- | -- | -- |
+| pp | String | (proxy) payment pointer | "\$spsp.example.com/alice" |
+| pt | String | plain text | "Hello World" |
```
sharafian, line 39: we should determine the largest payload we can encrypt/decrypt in the time limit that workers allows. If we find it to be restrictive at all I think we should consider the approach of returning keys instead of passing plaintext/ciphertext to the server directly

sharafian (changes requested):
Got some stuff on the README that we can discuss, then I can take a pass over the code after that. Overall seems very cool, it's awesome we can implement this with such simple code

workers/exclusive-content/LICENSE_MIT
sabinebertram: No problem, I'll change that. I think this was the default.

workers/exclusive-content/README.md
sabinebertram: It is per worker

workers/exclusive-content/README.md
sabinebertram: Yes, I'll change it. It was just easier for all the testing because I didn't have to type that much ;-)

workers/exclusive-content/README.md
sabinebertram: It is an array buffer that was converted to a utf-8 string and then base64 encoded. How detailed does it have to be?

workers/exclusive-content/README.md
sabinebertram: Yes, you are right. I prefer option 2 since that way everybody can use their own verifier. But for a first implementation, we could also use one verifier for everybode (the new service). 
If we are moving encryption and decryption of the plaintext out of the worker, the verifier decryption would still happen inside the worker but I guess that is ok.

workers/exclusive-content/README.md
sabinebertram: Yes. I'll change that as soon as the new verifier version is out. Right now, I'm still developing against an old version.

workers/exclusive-content/README.md
sabinebertram: I think we have decided to move encryption and decryption outside of the worker. That way, the original exclusive content never leaves the content creator's browser.

workers/exclusive-content/README.md
sharafian: if we say that it's base64 encoded then that's fine; it just wasn't clear as the iv appears to be all numbers

workers/exclusive-content/README.md
sharafian: yeah if we're not passing the ciphertext to the server it's a little tricky. For this version the most reasonable solution is probably just to configure a verifier endpoint as a variable for the worker.

Another option would be to have a JWT-like signed token that the exclusive content generator produces at encryption time and takes in at decryption time. But maybe that's a feature for later if we can get away with having one verifier for now

workers/exclusive-content/README.md
sabinebertram: Oh sorry, I didn't look at this carefully. The iv is just a random number, converted to string. I can make that alphanumeric like the nonce, too.

workers/exclusive-content/README.md
sabinebertram: I actually went with the flow of passing the encrypted verifier endpoint but I'm happy to change that.

workers/exclusive-content/index.js
```diff
@@ -0,0 +1,164 @@
...
+    bufView[i] = str.charCodeAt(i)
+  }
+  return buf
+}
+
+async function importMasterKey() {
+  const encoder = new TextEncoder()
+  return crypto.subtle.importKey(
+    'raw',
+    encoder.encode(MASTERKEY),
```
sharafian, line 124: you could use your encode helper function here instead of making a text encoder

workers/exclusive-content/index.js
```diff
@@ -0,0 +1,164 @@
...
+  } else if (request.method === 'OPTIONS') {
+    return new Response('OK', { headers })
+  } else {
+    return new Response('Not found', { headers, status: 404 })
+  }
+}
+
+async function handleNewKeyDerivation(body) {
+  const nonce = getRandomValue()
+  const keyBuffer = await deriveKey(encode(body.paymentPointer + nonce))
```
sharafian, line 67: making the deriveKey method just take 2 arguments (paymentPointer, nonce) and do the concatenation + encoding inside could save a little duplication

workers/exclusive-content/index.js
```diff
@@ -0,0 +1,164 @@
...
+}
+
+async function handleExKeyDerivation(body) {
+  const keyBuffer = await deriveKey(encode(body.paymentPointer + body.nonce))
+  const verifier = await decrypt(
+    str2ab(atob(body.encVerifier)),
+    keyBuffer,
+    encode(body.initVector)
+  )
+  const payment = await verifyReceipt(decode(verifier), body.receipt)
```
sharafian, line 81: nit: as payment is a boolean here the name `verified` might be more appropriate

sharafian (changes requested):
Looks really good! Most of my comments are very minor. Only major thing is that this should be rewritten to typescript

workers/exclusive-content/index.js
```diff
@@ -0,0 +1,164 @@
+/*
```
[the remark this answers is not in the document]
sabinebertram: I implemented the minor changes, now I'm on it.

sharafian (changes requested):

workers/exclusive-content/README.md
```diff
@@ -66,9 +66,9 @@ JSON object containing
 | Name | Type | Description | Example |
 | -- | -- | -- | -- |
 | paymentPointer | String | | "\$spsp.example.com/alice" |
-| nonce | String | 16 random alphanumeric characters | "ea22smqdrugcc8ty" |
+| nonce | String | base64-encoded random values generated with [crypto.getRandomValues()](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues) | "iVBojzgBQiyIlaK2yzKUNQ==" |
 | encVerifier | String | base64-encoded AES-GCM encryption of the verifier endpoint | "nsNBmLVm11WDVbst+jJuxkA17vEH+/W6yIPeQdnoK3mdT8D/LA==" |
-| initVector | String | random initialization vector used for `encVerifier` decryption, 16 random alphanumeric characters | "6pcvwv40ckftubz8" |
+| initVector | String | base64-encoded random values generated with [crypto.getRandomValues()](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues) | "A8S8GwC/luARiJG6xCNQBw==" |
```
sharafian, line 71: nit: I'd keep the note about this being used for encVerifier decryption

workers/exclusive-content/src/lib/requestHandlers.ts
```diff
@@ -9,14 +9,16 @@ import {
 } from './helpers'
 import { verifyReceipt } from './receiptVerifier'
 
+headers.set('Content-Type', 'application/json;charset=UTF-8')
```
sharafian, line 12: mutating this global is a little weird. Is this change meant to affect the object everywhere?

sharafian (changes requested):
looks good, just a really small thing left

workers/exclusive-content/README.md
sabinebertram: oops, that got lost in copy pasting. Thanks.

sharafian (approved):

sabinebertram:
👆 I didn't have git.config set up after upgrading my OS, hence the force-push after rebasing and adding the user information.

sharafian (approved):

PR #135, opened 2020-10-23
Adding Infinity Search to the list of search engines 

InfinitySearch:
Our search engine is Web Monetization enabled and it doesn't display ads to Coil members. This just adds us to list of these search engines next to Runnaroo.


adrianhopebailie (approved):

adrianhopebailie:

PR #141, opened 2020-11-17
docs: added page for receipt verifier service

melissahenderson:


docs/receipt-verifier.md
```diff
@@ -0,0 +1,98 @@
...
+npm start
+```
+
+## Update your Web Monetization meta tag
+
+Typically, a Web Monetization meta tag looks something like this:
+
+`<meta name=“monetization" content="$wallet.example.com/alice">`
+
+Update your meta tag so that your URL-encoded payment pointer is in the path of the receipt verifier service’s URL.
```
sharafian, line 48: We should link somewhere like https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent to give people a little help on URI-encoding their payment pointer

docs/receipt-verifier.md
```diff
@@ -0,0 +1,98 @@
...
+And your receipt verifier service’s URL is `https://receipt-verifier.com`, then you’ll set your meta tag’s `content` to either of the following:
+
+* `https://receipt-verifier.com/%24webmonetization.org%2Fapi%2Frevshare%2Fpay%2FW1siJHdhbGxldC5leGFtcGxlL2FsaWNlIiwxMCwiQWxpY2UiXSxbIiR3YWxsZXQuZXhhbXBsZS9ib2IiLDEwLCJCb2IiXV0`
+* `$receipt-verifier.com/ %24webmonetization.org%2Fapi%2Frevshare%2Fpay%2FW1siJHdhbGxldC5leGFtcGxlL2FsaWNlIiwxMCwiQWxpY2UiXSxbIiR3YWxsZXQuZXhhbXBsZS9ib2IiLDEwLCJCb2IiXV0`
+
+## Set up a monetizationprogress event listener
+
+When a WM sender receives a receipt, the sender notifies the user agent of the successful payment by submitting a `monetizationprogress` event containing a receipt.
+
+Add the following client-side code to your website to listen for a `monetizationprogress` event containing a `receipt`. When a receipt exists, your website will send it to the receipt verifier in the event listener to confirm the payment.
```
sharafian, line 74: we should rethink this example; the security isn't actually improved by speaking directly to the receipt verifier from the frontend, because anyone could change what code is on the frontend or intercept the request to make it look like the request succeeded. The proper way to use the receipt verifier would be to submit a receipt to your app's backend and then have your app's backend submit the receipt to the verifier in order to ensure it's validity. Then the app backend can securely take action like returning exclusive content or giving some perk to the user's account.

sharafian (changes requested):

docs/receipt-verifier.md
```diff
@@ -0,0 +1,118 @@
...
+id: receipt-verifier
+title: Receipt Verifier Service
+sidebar_label: Receipt verifier service
+---
+
+This page explains how to set up a receipt verifier service.
+
+As more creators adopt Web Monetization to serve premium and/or ad-free content, there’s a greater need for reliable payment verification. Web Monetization browser events, like `monetizationstart`, are helpful indicators of payment, but can be spoofed by savvy users to access exclusive content without paying.
+
+[Interledger STREAM receipts](https://interledger.org/rfcs/0039-stream-receipts/) provide payment recipients (such as web monetized site owners) with verifiable proofs of payment. As a payment recipient, you can verify STREAM receipts yourself by setting up your own verification service or by using a third-party service. You don't need to run your own Interledger connector or SPSP server to verify STREAM receipts.
```
sharafian, line 11: we should use the term 'receipt verifier service' consistently, so "you can verify STREAM receipts yourself by setting up your own receipt verifier service"

docs/receipt-verifier.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Service overview
+
+The receipt verifier service acts as a proxy between the Web Monetization sender and receiver.
+
+Before a payment is made, the WM sender sends a query to the WM receiver. The query passes through the verifier service, which adds a receipt secret and receipt nonce.
+
+When the WM receiver receives money from the payment, it generates a receipt (which includes a signature generated using the receipt secret) and sends the receipt to the WM sender.
+
+The WM sender passes the receipt to the verifier service so the verifier can confirm the payment (as only the WM receiver and verifier service know the receipt secret). The verifier must verify the receipt before accepting the amount as paid.
```
sharafian, line 25: this portion is implemented by the site, but this paragraph makes it sound a bit like it happens automatically. The site will have to take the receipt from the web monetization events and then submit it to its backend (where the backend can submit the receipt to the receipt verifier service).

docs/receipt-verifier.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Install the receipt verifier service package
+
+You can get the [receipt verifier service package](https://github.com/coilhq/receipt-verifier) or run the verifier as a [Kubernetes service](https://github.com/coilhq/receipt-verifier/tree/main/config/base).
+
+**Example**
+
+This example shows Coil's implementation of the receipt verifier package, which requires Redis. You can write a different implementation that doesn't rely on Redis.
+
+```
```
sharafian, line 35: we might want to include the git clone here to make the example more complete

```
git clone https://github.com/coilhq/receipt-verifier.git
cd receipt-verifier
```

docs/receipt-verifier.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Update your Web Monetization meta tag
+
+Typically, a Web Monetization meta tag looks something like this:
+
+`<meta name=“monetization" content="$wallet.example.com/alice">`
+
+For queries from the WM sender to be proxied by the receipt verifier service to your payment pointer:
+
+1. URI-encode your payment pointer. See [encodeURIComponent()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent) on MDN Web Docs if you need help.
```
sharafian, line 50: will our site support MDX? https://v2.docusaurus.io/docs/markdown-features/#embedding-react-components-with-mdx I could make a little inline component for URI encoding pointers in the future.

docs/receipt-verifier.md
```diff
@@ -0,0 +1,118 @@
...
+```html
+<head>
+  <!-- This should be set to .... -->
+  <meta name="monetization" content="https://receipt-verifier.com/%24wallet.example.com%2Falice">
+  <script>
+    if (document.monetization) {
+        document.monetization.addEventListener('monetizationprogress', event => {
+            // A payment has been received
+
+            // Connect to your site’s backend to validate the payment
```
sharafian, line 88: let's add a note like "this _does not_ connect directly to the receipt verifier" to make it even clearer. I feel like this is something people could get tripped up on

docs/receipt-verifier.md
```diff
@@ -0,0 +1,118 @@
...
+    }
+  </script>
+</head>
+```
+
+Your backend can send the receipt to the receipt verifier’s `/verify` endpoint. Here’s an example for an Express.js server:
+
+```javascript
+app.post('/verifyReceipt', async (req, res) => {
+  const resp = await fetch('https://receipt-verifier.com/verify', {
```
sharafian, line 108: unless we own receipt-verifier.com it's better to not use a real domain in an example. `receipt-verifier.example` would be safer

sharafian (changes requested):
just have a few clarifications we should make

docs/receipt-verifier.md
melissahenderson: @sharafian How about this:

The site takes the receipt from the Web Monetization events, then submits the receipt to its backend. The backend then submits the receipt to the receipt verifier service so the service can confirm the payment (as only the WM receiver and the receipt verifier service know the receipt secret). The service must verify the receipt before accepting the amount as paid.

docs/receipt-verifier.md
melissahenderson: I believe the site will support that. The example block Docusaurus provides works locally, as do .mdx files. I also imported a component from Rebass (https://mdxjs.com/getting-started/#imports), and that was OK too.

sharafian (approved):
Looks good!

We now have a publicly available verifier too, with the URL of `https://webmonetization.org/api/receipts`. We can merge this and then make that addition as a subsequent PR, to keep it manageable

PR #142, opened 2020-11-20
feat: fix up exclusive content worker to make publishable

sharafian:
just did a few fixes so that 'wrangler publish' would work. I've published the worker in it's current state to webmonetization.org (the route is `https://webmonetization.org/api/exclusive-content/*` as written in wrangler.toml) so that you can more easily test the frontend.

### DIFF
--- a/workers/exclusive-content/.gitignore
+++ b/workers/exclusive-content/.gitignore
@@ -1,3 +1,2 @@
 /dist
 /node_modules
-wrangler.toml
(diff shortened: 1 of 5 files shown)


sabinebertram:
Since you deleted the `wrangler.toml.example`, should we include the real `wrangler.toml`?

sabinebertram (approved):
Thanks for making these changes!

PR #146, opened 2020-12-09
docs: add info about publicly available verifier

melissahenderson:
Decided to leave the monetizationprogress event listener section as-is


wilsonianb (approved):
:+1:

sharafian (approved):

PR #148, opened 2020-12-14
feat: pass on origin

sharafian:
passes the origin along to the receipt verifier so that it returns the CORS headers


wilsonianb (approved):
:+1: 
Deployed and tested locally
Submitting receipts to `'http://127.0.0.1:8787/api/receipts/verify` from the browser failed without this change due to CORS and succeeded with the change.

PR #155, opened 2021-01-06
doc: investigate broken links

melissahenderson:
I'm finding that some links break for me, and when I go back from the error page, the link then works properly. I've updated two pages to try and solve the issue.


adrianhopebailie (approved):
LGTM

PR #161, opened 2021-02-05
feat: use google cloudrun verifier to avoid worker to worker call 

sabinebertram:
Since a cloudflare worker cannot call another cloudflare worker, we change the verifier URL to the google cloudrun URL to talk to the verifier directly and not via the cloudflare worker. 

This needs to be deployed to cloudflare after merging.


wilsonianb (approved):
:+1:

sharafian (approved):

PR #194, opened 2021-08-10
chore: enable pr_preview for spec

marcoscaceres:
Allows us to preview pull requests sent to the spec. 

http://github.com/tobie/pr-preview

(we use it a lot at the W3C)


AlexLakatos (approved):
LGTM 🥨

PR #202, opened 2021-09-07
Chore: enable auto-tidy of spec on PRs

marcoscaceres:
Makes it easier for folks to contribute, without worrying about worrying about whitespace. Checks HTML for us too.  

This will make it easier to review all other PRs too.


AlexLakatos (approved):
LGTM 🎫

PR #205, opened 2021-09-07
Editorial: rewrite terminology section

marcoscaceres:


static/specification.html
```diff
@@ -98,35 +98,29 @@ <h2>
...
         <dt>
           <dfn>Web Monetization receiver</dfn>:
         </dt>
         <dd>
-          the party receiving payments on behalf of the website.
+          The party receiving payments on behalf of the website.
         </dd>
         <dt>
-          <dfn>User agent</dfn>:
+          <dfn>Payment end-point</dfn>:
```
AlexLakatos, line 119: This is what we're calling both the "pointer" and the "href"? Should we keep this as the href, and add a separate entry for the payment pointer?

marcoscaceres: Agree... although, I think this one is ok, as the spec explicitly states that the `href` must be a "payment end-point" (which is a SPSP end-point).

AlexLakatos (approved):
LGTM 🦖

static/specification.html
AlexLakatos: Great!

PR #193, opened 2021-08-10
New Web Monetization Specification Proposal

marcoscaceres:
A preview of the new proposed specification can be viewed at https://webmonetization-preview.netlify.app/

Removed:
  * using meta elements
  * various conformance restrictions 
  * flow part... will be part of link relationship instead
  * state entirely
  * document.monetization
 
Added: 

* Allow multiple link elements - but the WM agent can limit how many connections it handles.
* The "progress" event is now called "monetization", and is fired at the <link> element .
* introduce a new global event handler attribute: `onmonetization`, which is on every HTML element.
* We now route all events through the <link> element.
* Developer-recoverable errors are handled via the `<link onerror>` event handler.
* Successfully loading/processing SPSP is reflected via the `<link onload>` event handler.
* Defined CSP `monetization-src`. 
* assetScale in now an octet (uint8).

Closes #144
Closes #149
Closes #159  
Closes #19
Closes #192
Closes #20
Closes #201
Closes #21 
Closes #214
Closes #215
Closes #218
Closes #232
Closes #22 
Closes #233
Closes #28
Closes #29
Closes #41
Closes #23
Closes #216
Closes #220 
Closes #165 
Closes #204 
Closes #219  
Closes #237

PR #208, opened 2021-09-07
BREAKING CHANGE: use Navigator interface instead of Document

### LINKED ISSUE #24: `document.monetization` vs global
From web-monetization created by [marcoscaceres](https://github.com/marcoscaceres): adrianhopebailie/web-monetization#23  

Wondering what the rationale was for attaching monetization to the `Document` object? It doesn't really feel like the right place to attach this, as monetization doesn't itself affect the document or have anything specifically to do with the document interface. We should consider maybe having monetization either on Navigator or its own `Monetization` namespace.

marcoscaceres:
Closes #24

### DIFF
--- a/static/specification.html
+++ b/static/specification.html
@@ -150,10 +150,10 @@
     </section>
     <section>
       <h2>
-        Extensions to the `Document` interface
+        Extensions to the `Navigator` interface
       </h2>
       <pre class="idl">
-        partial interface Document {
+        partial interface Navigator {
           readonly attribute Monetization monetization;
         };
       </pre>
@@ -236,7 +236,7 @@
             with a non-zero amount, the provider notifies the browser:
               <ul>
                 <li>The <a>user agent</a> sets
-                <code>document.monetization.state</code> to
+                <code>navigator.monetization.state</code> to
                 <code>started</code>.
                 </li>
                 <li>This MUST occur before the `start` event is fired.
@@ -245,7 +245,7 @@
                 Payment Setup Endpoint and the Monetization ID as properties.
                 </li>
                 <li>The <a>user agent</a> also emits a {{ progress }} event
-                from <code>document.monetization</code>, corresponding to this
+                from <code>navigator.monetization</code>, corresponding to this
                 first packet.
                 </li>
               </ul>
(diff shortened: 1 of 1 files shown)


AlexLakatos (approved):
This seems to be one of the most contended issues. Syntactically, I agree with it, Monetization should live on the Navigator instead of the Document. Window is just a "couldn't come up with anything else" place, and prone to clashing with user code.

I wonder how this will clash with the existing Web Monetization provider extension implementation. @sharafian any strong opinions against this?

sharafian:
using navigator is alright with me, assuming that there aren't any quirks that come up in the implementation. It's also nice because we can keep the document interface around until enough users have migrated off of it, and implement these new breaking changes on the navigator interface

AlexLakatos:
Perfect! That also makes #209 easier to approve then!

static/specification.html
```diff
@@ -236,7 +237,7 @@ <h2>
             with a non-zero amount, the provider notifies the browser:
               <ul>
                 <li>The <a>user agent</a> sets
-                <code>document.monetization.state</code> to
+                <code>navigator.monetization.state</code> to
```
sublimator, line 240: What states do we imagine this having and how will they be used in if we are collapsing the events, and in particular removing the stop event? 

Currently we have `pending | started | stopped`
We'll set the state to `stopped` but not have a corresponding event ?
We'll no longer have a `pending` event but set the `state` ?

sublimator: This reminds me of a conversation we had internally on slack with Matt Baer from write.as 
He wasn't happy just using a timeout to determine a user wasn't willing or able to monetize

marcoscaceres: Hey, so we can add as many events as we need. The collapsing of events is just to merge the interface, but it doesn't need to limit the number of states we can have. 

>  We'll set the state to stopped but not have a corresponding event ?

No, let me clarify with an example 

```
navigator.monetization.state; // changes to "stopped"
navigator.monetization.onstatechage = event => {
     switch(navigator.monetization.state):
            case "stopped":
                 // do stuff..
            case "started":
                 // do stuff..
            case "whatever-we-need":
                // do stuff..
            default: 
                // do default stuff..
}
```

sublimator: Got it! 
In the latest extension, I mapped pending/started to the proposed 'interactive', and stopped to 'idle'
Would be good if we can agree on some terms @ https://github.com/WICG/webmonetization/issues/220

I'm pretty sure @sharafian would like to see as little change as possible

marcoscaceres:
Duplicate of #193

PR #229, opened 2021-11-09
chore: update algolia api key with the change to it being called crawler

traviscrist:
Updates Algolia search to the new API key and APP ID which is needed for their move to Algolia Crawler

Screen Shot 2021-11-09 at 9 40 56 AM


traviscrist:
This key is not for webmon its for coil.com docs. Closing

PR #235, opened 2021-11-29
Add permissions policy integration

### LINKED ISSUE #28: Define feature policy for monetization
From web-monetization created by [jpettitt](https://github.com/jpettitt): adrianhopebailie/web-monetization#26  

The spec as drafted has multiple issues related to revenue sharing, fraud, and abuse prevention.

#16 and #23 have some discussion of where the `<meta>` tag and `monetization` object should live.   The `document` vs `navigator` choice has some implications for abuse prevention.

As written the spec disallows `meta` tags in iframes.  This is impossible to enforce as `<iframe>` contents are not visible to parent frames.  Depending on the, as yet unspecified error handling, an iframe could include a monetization `meta` tag and trigger a denial of service on the parent by virtue of having more than one tag.

Conversely, if `meta` tags in `iframes` are ignored a malicious actor could frame content pages and include a monetization tag in the top frame thus hijacking the payment from the child frame content.  This become particularly tricky when dealing with embedded videos where there may be multiple, legitimately monetized child frames on a page.

The mechanics of allocating payment between multiple claimants, some legitimate some not are extremely complex.  For example a page that collates the best cat videos from YouTube probably deserves some revenue for bringing the audience.  While at the same time the embedded videos also deserve revenue to reward the creators.  This is currently achieved vi banner and pre roll ads respectively.  It's unclear how this can be done within the current spec.

### LINKED ISSUE #214: Add permissions policy integration
The spec is currently somewhat unclear about the role of iframe and how web monetization works with third-party contexts.

It should be possible for embedded iframes to be monetized, but also restriction monetization to first-party iframes. Then we can just use permissions policy to enable payments for third-parties.

In other words/html: 

```html
<iframe allow="monetization" src="https://remote.com"></iframe>
```

marcoscaceres:
Closes #214 
Closes #28

### DIFF
--- a/static/specification.html
+++ b/static/specification.html
@@ -704,6 +704,22 @@
           </section>
         </section>
       </section>
+      <section id="permissions-policy" data-cite="permissions-policy">
+        <h2>Permissions Policy integration</h2>
+        <p>
+          This specification defines a [=policy-controlled feature=] identified
+          by the string <dfn class="permission">"monetization"</dfn>". Its
+          <a>default allowlist</a> is `'self'`.
+        </p>
+        <aside class="note">
+          <p>
+            A <a>document</a>’s [=Document/permissions policy=] determines
+            whether [^link^] elements are monetized - particularly in
+            third-party content, such as ifr. When disabled, monetization is
+            disabled and related events won't be dispatched.
+          </p>
+        </aside>
+      </section>
       <section id="http-headers">
         <h2>
           HTTP Headers
--- a/static/specification.html
+++ b/static/specification.html
@@ -705,7 +705,9 @@
         </section>
       </section>
       <section id="permissions-policy" data-cite="permissions-policy">
-        <h2>Permissions Policy integration</h2>
+        <h2>
+          Permissions Policy integration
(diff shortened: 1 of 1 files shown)


static/specification.html
```diff
@@ -704,6 +704,22 @@ <h3>
...
+        <p>
+          This specification defines a [=policy-controlled feature=] identified
+          by the string <dfn class="permission">"monetization"</dfn>". Its
+          <a>default allowlist</a> is `'self'`.
+        </p>
+        <aside class="note">
+          <p>
+            A <a>document</a>’s [=Document/permissions policy=] determines
+            whether [^link^] elements are monetized - particularly in
+            third-party content, such as ifr. When disabled, monetization is
```
sublimator, line 718: "such as ifr" -> "such as in iframes" ? 
It's obvious what you mean, but ??

marcoscaceres: oops! good catch.

AlexLakatos (approved):
LGTM.

marcoscaceres:
Now part of  #193

PR #265, opened 2022-03-01
fix: some minor edits

sublimator:


static/specification.html
```diff
@@ -302,7 +302,7 @@ <h2>
           <dfn>Payment end-point</dfn>:
         </dt>
         <dd>
-          A [=URL=] to a <a data-cite="SPSP#specification">SPSP end-point</a>
+          An [=URL=] to a <a data-cite="SPSP#specification">SPSP end-point</a>
```
marcoscaceres, line 305: See "a URL" is correct as per https://url.spec.whatwg.org
```suggestion
          A [=URL=] to a <a data-cite="SPSP#specification">SPSP end-point</a>
```

marcoscaceres (approved):
Couple of nits/suggestion! thanks @sublimator!!!

static/specification.html
```diff
@@ -693,7 +693,7 @@ <h2>
           <dfn>paymentPointer</dfn> attribute
         </h2>
         <p>
-          A URL representing a [=payment end-point=]. When getting, returns the
+          An URL representing a [=payment end-point=]. When getting, returns the
```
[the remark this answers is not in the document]
sublimator: I guess you read it as `y`oo-are-ell with a consonant leading sound

static/specification.html
sublimator: Yeah, I guess so, I'd stomach it better if it was spelled U.R.L :)

PR #273, opened 2022-04-20
revert: "Updates to support Open Payments"

sublimator:
Reverts WICG/webmonetization#269

The only approval on the PR was from a member of the WM working group that recently retired: https://github.com/WICG/webmonetization/pull/193/commits/1e859d1316ec2925d630cb854ca07a6120f1ee12

### DIFF
--- a/static/specification.html
+++ b/static/specification.html
@@ -59,6 +59,10 @@
             title: 'Interledger Protocol',
             href: 'https://interledger.org/rfcs/0027-interledger-protocol-4/',
           },
+          SPSP: {
+            title: 'Simple Payments Setup Protocol',
+            href: 'https://interledger.org/rfcs/0009-simple-payment-setup-protocol',
+          },
           STREAM: {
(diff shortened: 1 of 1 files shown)


adrianhopebailie:
I don't think that's a reason to revert the PR. This is still the direction we want to go. That PR is not against the spec it is against the v2 PR.

sublimator:
> This is still the direction we want to go.
Who is we? It's the direction you (and to some extent I) want to go, but it's hardly a majority. So the fact that the PR got only one approval from a retired editor is relevant. 

The last WM meeting I attended the general consensus was that they didn't want to restrict things too much and wanted to keep STREAM/SPSP around as options. @sabineschaller and @AlexLakatos were on the same page there. That was still the case as of today. 

I understand your possible frustration on this matter. I apologize that I haven't gotten around to doing an OP2 proof-of-concept with the extension. It was quite a high priority for me but I have had a really bad time health wise recently.

PR #276, opened 2022-05-09
chore: remove non W3C license

sabineschaller:
The repo currently has 2 licenses: 
* https://github.com/WICG/webmonetization/blob/main/LICENSE.md -> W3C
* https://github.com/WICG/webmonetization/blob/main/LICENSE -> Apache 2.0

Since the WICG always operates under the W3C license, [LICENSE](https://github.com/WICG/webmonetization/blob/main/LICENSE) should be removed.


AlexLakatos (approved):
LGTM 📎 

Thanks for the PR!

PR #284, opened 2022-10-06
Update w3c.json

marcoscaceres:
I probably shouldn't be the contact here anymore... @dontcallmedom, can you suggest someone else?


dontcallmedom:
since this is a WICG spec, I believe it would conventionally be one of the WICG chairs?

sublimator:
@adrianhopebailie 

What about you?

AlexLakatos:
@sublimator fixed this in #308

PR #285, opened 2022-10-12
fix: various

adrianhopebailie:
I have tried to address some of the discussion in the main link_rel PR. I think that this helps to make the whole spec a little cleaner.

The goal is that the WM spec should not have too much detail about ILP or Open Payments as these are evolving and leaking details into this spec makes changes hard.

fix: math in examples
As pointed out by @sublimator 

fix: payment pointer (not endpoint)
Replace "payment end-point" with payment pointer

fix: send amount in event
Fix ambiguity about send vs receive amount and add warning about changes

fix: removed some leaking external protocol refs
Adjust some wording to remove ambiguity about Open Payments concepts.

fix: former editors, editor details
Move former editors into former editors block

### DIFF
--- a/static/specification.html
+++ b/static/specification.html
@@ -18,12 +18,6 @@
         },
         editors: [
           {
-            name: 'Marcos Caceres',
-            company: 'W3C',
-            companyURL: 'https://w3.org/',
-            retiredDate: '2022-03-23',
-          },
-          {
             name: 'Alex Lakatos',
             email: 'alex@interledger.org',
             company: 'Interledger Foundation',
@@ -31,11 +25,18 @@
           },
           {
             name: 'Adrian Hope-Bailie',
-            email: 'adrian@coil.com',
+            email: 'adrian@fynbos.dev.com',
+            company: 'Fynbos Technologies Limited',
+            companyURL: 'https://fynbos.dev',
+          },
+          {
+            name: 'Nicholas Dudfield',
+            email: 'niq@coil.com',
             company: 'Coil Technologies Inc.',
             companyURL: 'https://coil.com',
-            retiredDate: '2021-12-01',
           },
+        ],
+        formerEditors: [
           {
             name: 'Ben Sharafian',
             email: 'ben@coil.com',
@@ -44,11 +45,11 @@
             retiredDate: '2022-02-01',
           },
           {
-            name: 'Nicholas Dudfield',
-            email: 'niq@coil.com',
-            company: 'Coil Technologies Inc.',
-            companyURL: 'https://coil.com',
-          },
+            name: 'Marcos Caceres',
+            company: 'W3C',
+            companyURL: 'https://w3.org/',
+            retiredDate: '2022-03-23',
+          }
         ],
         shortName: 'web-monetization',
         specStatus: 'CG-DRAFT',
@@ -193,7 +194,7 @@
               link.addEventListener("monetization", event =&gt; {
                 // See how much was sent and in what currency
                 const { amount, assetCode, assetScale } = event;
-                console.log(`Browser sent ${assetCode}${amount / (10 * assetScale)}.`);
+                console.log(`Browser sent ${assetCode}${amount / (10 ** assetScale)}.`);
 
                 // Use the incoming payment URL to verify the payment at the receiver.
                 verifyPayment(event);
@@ -244,7 +245,7 @@
                 // The incoming payment was fetched successfully
                 const { receivedAmount } = JSON.parse(response.json())
                 const { amount, assetCode, assetScale } = receivedAmount;
-                console.log(`Received ${assetCode}${amount / (10 * assetScale)}.`);
+                console.log(`Received ${assetCode}${amount / (10 ** assetScale)}.`);
               }
             }
           </pre>
@@ -256,7 +257,7 @@
         </h3>
         <p>
           The following example shows how to monetize various types of media
-          using different [=payment end-points=].
+          using different [=payment pointers=].
         </p>
         <aside class="example">
           <pre class="html">
@@ -284,7 +285,7 @@
           <dfn data-lt="monetized">Monetization</dfn>:
         </dt>
         <dd>
-          A payment made by a user to a website, facilitated through a
+          Payments made by a user to a website, facilitated through a
           [=monetization provider=].
         </dd>
         <dt>
@@ -301,14 +302,14 @@
         </dt>
         <dd>
           The party receiving payments on behalf of the website, whose details
-          are provided by a [=payment end-point=].
+          are provided by a [=payment pointer=].
         </dd>
         <dt>
-          <dfn>Payment end-point</dfn>:
+          <dfn>Payment Pointer</dfn>:
         </dt>
         <dd>
-          A [=URL=] to a <a data-cite="Open Payments#specification">Open
-          Payments account end-point</a> (i.e., a JSON resource containing
+          A [=URL=] to an <a data-cite="Open Payments#specification">Open
+          Payments API entry-point</a> (i.e., a JSON resource containing
           details that facilitate payment to an account).
         </dd>
         <dt>
@@ -316,9 +317,9 @@
         </dt>
         <dd>
           An session between a [=monetization provider=] and [=monetization
-          receiver=] initiated by the creation of an incoming payment resource
-          at the [=monetization receiver=]. One or more payments can be sent by
-          the [=monetization provider=] in a single session.
+          receiver=] initiated by the [=monetization provider=] at the
+          [=monetization receiver=]. One or more payments can be sent by the
+          [=monetization provider=] in a single session.
         </dd>
       </dl>
     </section>
@@ -365,7 +366,7 @@
         </p>
       </aside>
       <p>
-        The monetization keyword indicates a [=payment end-point=] used to
+        The monetization keyword indicates a [=payment pointer=] used to
         monetize the document.
       </p>
       <p>
@@ -581,8 +582,8 @@
         monetization event
       </h2>
       <p>
-        When the [=payment session=] has sent an |outgoing payment| with a
-        non-zero amount, perform the following steps:
+        When the [=payment session=] has sent a |payment| with a non-zero
+        amount, perform the following steps:
       </p>
       <ol class="algorithm">
         <li>Let |target:HTMLLinkElement| be the {{HTMLLinkElement}} associated
@@ -592,7 +593,7 @@
         </li>
         <li>Let |eventInitDict:MonetizationEventInit| be a
         {{MonetizationEventInit}} dictionary, whose members are initialized to
-        match |outgoing payment|'s details.
+        match |payment|'s details.
           <p class="issue">
             We probably need to canonicalize things like `assetCode` to
             uppercase. See {{PaymentCurrencyAmount/currency}} for how it's done
@@ -661,9 +662,19 @@
           <dfn>amount</dfn> attribute
         </h2>
         <p>
-          The amount delivered by an ILP packet. When getting, returns the
-          value it was initialized with.
+          The amount <em>sent</em>. When getting, returns the value it was
+          initialized with.
         </p>
+        <div class="warning">
+          <p>
+            The `amount` attribute previously reflected the amount
+            <em>received</em>.
+          </p>
+          <p>
+            To determine the amount received use either the `receipt` or
+            `incomingPayment` attribute.
+          </p>
+        </div>
       </section>
       <section>
         <h2>
@@ -702,8 +713,8 @@
         <p>
           `null` or a base64-encoded [[[Receipt]]] issued by the [=monetization
           receiver=] to the [=monetization provider=] as proof of the total
-          amount received in the stream. When getting, returns the value it was
-          initialized with.
+          amount received in the [=payment session=]. When getting, returns the
+          value it was initialized with.
         </p>
       </section>
       <section>
@@ -711,7 +722,7 @@
           <dfn>paymentPointer</dfn> attribute
         </h2>
         <p>
-          A [=URL=] representing a [=payment end-point=]. When getting, returns
+          A [=URL=] representing a [=payment pointer=]. When getting, returns
           the value it was initialized with.
         </p>
       </section>
@@ -757,7 +768,7 @@
         </h3>
         <p>
           The monetization-src directive restricts the URLs from which a
-          [=payment end-point=] is loaded. The syntax for the directive's name
+          [=payment pointer=] is loaded. The syntax for the directive's name
           and value is described by the following ABNF:
         </p>
         <pre class="abnf">
@@ -845,17 +856,17 @@
         or start monetization of a particular site if they wish to do so).
       </p>
       <p>
-        As [=payment end-points=] are generally provided as a service (e.g.,
+        As [=payment pointers=] are generally provided as a service (e.g.,
         Uphold), a <abbr title="Cross Site Scripting">XSS</abbr> attack could
-        inject malicious [=payment end-points=] into a page that uses the same
+        inject malicious [=payment pointers=] into a page that uses the same
         service. To mitigate such an attack, it is RECOMMENDED that developers:
       </p>
       <ul>
-        <li>host a [=payment end-point=] on their own servers and proxy [[Open
+        <li>host a [=payment pointer=] on their own servers and proxy [[Open
         Payments]] requests on the server as needed.
         </li>
         <li>Set the `monetization-src` CSP directive to restrict requests to
-        origins they control and trust to host [=payment end-points=].
+        origins they control and trust to host [=payment pointers=].
         </li>
       </ul>
     </section>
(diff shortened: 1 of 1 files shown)


adrianhopebailie:
@sublimator - I think that this part is key to the whole separation of concerns:

> Otherwise, establish a new **payment session**.

This is where the browser (having successfully fetched the payment pointer) does processing and interfaces with the WM Provider to create a **payment session**.

The specifics of this are opaque to the website. If this session goes stale and is re-established later then there is no need for the website to know that.

static/specification.html
```diff
@@ -661,9 +662,19 @@ <h2>
           <dfn>amount</dfn> attribute
         </h2>
         <p>
-          The amount delivered by an ILP packet. When getting, returns the
-          value it was initialized with.
+          The amount <em>sent</em>. When getting, returns the value it was
```
sublimator, line 665: One of the points I was trying to make is that the field name itself should convey whether it's sent or delivered. 
i.e.  amount -> sentAmount

static/specification.html
```diff
@@ -661,9 +662,19 @@ <h2>
...
+          initialized with.
         </p>
+        <div class="warning">
+          <p>
+            The `amount` attribute previously reflected the amount
+            <em>received</em>.
```
sublimator, line 671: So by using sentAmount it means you could do things like an http request to initiate a payment on the backend, and rather than have to wait for the STREAM to finish, the backend could send to the incoming payment and respond immediately ?

static/specification.html
AlexLakatos: But we're not doing both amounts here, there's no ambiguity that needs disentangling. And by that logic, we'd have to use the "sent" prefix in front of the other properties, if we support both sent and received amounts. Same for received. I think keeping this property as "amount", and documenting it to indicate direction is a cleaner approach.

sublimator: > there's no ambiguity that needs disentangling.

It makes it very clear, documenting it in the field name. "amount" is much cuter/shorter than "sourceAmount" or "sentAmount" granted, but having seen bugs over the years where people forgot which is which (even Coil employees), I value things differently.

sublimator:
We should probably make a struct for amounts, probably matching OP (value,
not amount). I was liking PaymentAmountCurrency but if we are using OP ...

sublimator:
@adrianhopebailie 

> This is where the browser (having successfully fetched the payment pointer) does processing and interfaces with the WM Provider to create a payment session.
> The specifics of this are opaque to the website. If this session goes stale and is re-established later then there is no need for the website to know that.

Thanks for the clarification. By that you mean there is no need to emit another "load" event. 
My objection is with mandating that only the browser processes the payment pointer and the tradeoffs involved, but we can have that discussion elsewhere.

sublimator (approved):
For the record, I'm not totally convinced about sourceAmount only, but basically lgtm

static/specification.html
adrianhopebailie: Yes

static/specification.html
adrianhopebailie: I think `amountSent` makes sense.

We can deprecate `amount` and that way we avoid weird behaviour for old websites that still expect to find a received amount there.

`amount` would also be optional and new implementations would not set a value. I think it's better for old websites to break rather than assume this is a receive amount when its actually a sent amount.

adrianhopebailie:
> I was liking PaymentAmountCurrency but if we are using OP ...

Me too. I don't think the protocol used on the backend should be relevant we should use WebIDL types that already exist where we can.

static/specification.html
```diff
@@ -639,30 +635,50 @@ <h2>
...
         interface MonetizationEvent : Event {
           constructor(DOMString type, MonetizationEventInit eventInitDict);
-          readonly attribute DOMString amount;
-          readonly attribute DOMString assetCode;
-          readonly attribute octet assetScale;
+          readonly attribute DOMString? amount;
+          readonly attribute DOMString? assetCode;
+          readonly attribute octet? assetScale;
           readonly attribute DOMString? receipt;
+          readonly attribute PaymentCurrencyAmount amountSent;
```
sublimator, line 642: Note that the payments `PaymentCurrencyAmount` is a dictionary type, not interface, and thus not suitable for an event attribute. If we were to go down this route, we would have to define a new interface.

```
[SecureContext, Exposed=Window]
interface MonetizationAmount {
  readonly attribute DOMString value;
  readonly attribute DOMString currency;
};

[SecureContext, Exposed=Window]
interface MonetizationEvent : Event {
  constructor(DOMString type, MonetizationEventInit eventInitDict);
  readonly attribute MonetizationAmount amountSent;
```

While this would mean window.MonetizationAmount would be exposed, the constructor wouldn’t actually be. @sidvishnoi also says this is standard practice. There used to be a [NoInterfaceObject] decoration, but it was removed, and people just prefix the interfaces.

static/specification.html
sublimator: > Did you get the 5 million dong I sent??
Not yet.
What?? You said to send ASAP but you haven't even picked it up?
WU really gouged me!
Was like $295 USD!

What's the 'amountSent' there? 

I think "sent" by itself isn't all that clear.

static/specification.html
sublimator: Should there be any limitations on the amount of different assets "sent" in a given session ?  Just 1? 
Theoretically a provider could send using multiple source currencies, and it would complicate things a little having to check for that.  

Given we are going to use "monetization" event for tips, how does one know when a tip is sent? Once a value threshold has been crossed?

static/specification.html
adrianhopebailie: @marcoscaceres - you were close to the design of `PaymentCurrencyAmount `. WDYT?

Should we define a `MonetizationAmount` interface that aligns with the types of the `PaymentCurrencyAmount` dictionary?

adrianhopebailie: @sublimator - what if we simply have the following:

```
[SecureContext, Exposed=Window]
interface MonetizationEvent : Event {
  constructor(DOMString type, MonetizationEventInit eventInitDict);
  readonly attribute DOMString amountSent;
  readonly attribute DOMString amountSentCurrency;
```

At least these data types align with the `PaymentCurrencyAmount`

sublimator: and amountSentAssetCode I guess ?

sublimator: > There used to be a [NoInterfaceObject] decoration, but it was removed, and people just prefix the interfaces.

Sadly missed 

> what if we simply have the following

In practice that might not be too bad:
```
const {amountSent, amountSentCurrency} = event; 
```

adrianhopebailie: This feels marginally nicer...
` const { amountSent : { value, currency }} = event;`

sublimator: In 2022, using a compound type as an attribute on an event shouldn't really be this hard of a decision eh? WebIDL should be fixed. 

> This feels marginally nicer...
> const { amountSent : { value, currency }} = event;

You know you want it! Declare it so

adrianhopebailie: Done

adrianhopebailie:
@AlexLakatos - this is ready to merge into the new spec. WDYT?

static/specification.html
```diff
@@ -31,11 +25,18 @@
           },
           {
             name: 'Adrian Hope-Bailie',
-            email: 'adrian@coil.com',
+            email: 'adrian@fynbos.dev.com',
```
AlexLakatos, line 28:
```suggestion
            email: 'adrian@fynbos.dev',
```

static/specification.html
```diff
@@ -301,24 +302,24 @@ <h2>
...
         </dd>
         <dt>
           <dfn>Payment session</dfn>
         </dt>
         <dd>
           An session between a [=monetization provider=] and [=monetization
-          receiver=] initiated by the creation of an incoming payment resource
-          at the [=monetization receiver=]. One or more payments can be sent by
-          the [=monetization provider=] in a single session.
+          receiver=] initiated by the [=monetization provider=] at the
```
AlexLakatos, line 320:
```suggestion
          receiver=] initiated by the [=user agent=] at the
```

static/specification.html
```diff
@@ -301,24 +302,24 @@ <h2>
...
+          [=monetization receiver=]. One or more payments can be sent by the
+          [=monetization provider=] in a single session.
```
sabineschaller, line 322:
```suggestion
          [=monetization receiver=]. One or more payments can be initiated by the
          [=monetization provider=] in a single session.
```
The monetization provider could send it out of their account or out of the linked wallet's account.

static/specification.html
```diff
@@ -301,24 +302,24 @@ <h2>
         </dt>
         <dd>
           The party receiving payments on behalf of the website, whose details
-          are provided by a [=payment end-point=].
+          are provided by a [=payment pointer=].
         </dd>
         <dt>
-          <dfn>Payment end-point</dfn>:
+          <dfn>Payment Pointer</dfn>:
```
uchibeke, line 308: Given the explanation `A [=URL=] to an ...` and the use of `<link>` tag, could we use `Payment end-point` instead?

Also, for non-Interledger folks, `Payment Pointer` might be confusing and `Payment end-point` seems to be a better fit for a Web API

static/specification.html
```diff
@@ -631,6 +627,42 @@ <h2>
         </dd>
       </dl>
     </section>
+    <section data-dfn-for="MonetizationCurrencyAmount">
+      <h2>
+        `MonetizationCurrencyAmount` interface
+      </h2>
+      <p>
+        The `MonetizationCurrencyAmount` interface maps directly to the
+        {{PaymentCurrencyAmount}} dictionary as defined in [[payment-request]].
```
sabineschaller, line 636: Why define `MonetizationCurrencyAmount` if it **directly** maps to `PaymentCurrencyAmount`?

static/specification.html
```diff
@@ -639,35 +671,55 @@ <h2>
...
 
         dictionary MonetizationEventInit : EventInit {
-          required DOMString amount;
-          required DOMString assetCode;
-          required octet assetScale;
+          required DOMString? amount;
+          required DOMString? assetCode;
+          required octet? assetScale;
           required DOMString? receipt;
+          required PaymentCurrencyAmount amountSent;
```
sabineschaller, line 688: `PaymentCurrencyAmount` or `MonetizationCurrencyAmount`?

static/specification.html
```diff
@@ -677,42 +729,49 @@ <h2>
...
-          <dfn>assetScale</dfn> attribute
+          <dfn>assetScale</dfn> attribute (deprecated)
         </h2>
         <p>
           The scale of the amount. For example, USD would have an
           <code>assetScale</code> of <code>2</code> when denominated in cents.
           When getting, returns the value it was initialized with.
         </p>
       </section>
+      <div class="note" title="PaymentCurrencyAmount">
```
sabineschaller, line 740: `PaymentCurrencyAmount` or `MonetizationCurrencyAmount`?

sabineschaller (commented):
I'm not sure about 2 things:
1. switching back from payment-endpoint to payment pointer
2. Introducing `MonetizationCurrencyAmount` or `PaymentCurrencyAmount` and replacing `assetCode` and `assetScale`

static/specification.html
sublimator: https://github.com/WICG/webmonetization/pull/285#discussion_r995747033

sabineschaller: 💡

static/specification.html
adrianhopebailie: What is a **payment end-point**?

A payment pointer is a well-defined concept and this definition aligns exactly with what it is.

static/specification.html
adrianhopebailie: I believe this is still `PaymentCurrencyAmount` as it's a dictionary used to construct the event.

We may need someone with a deeper knowledge of browser internals like @sidvishnoi or @marcoscaceres to help here

adrianhopebailie:
@AlexLakatos @sabineschaller @uchibeke I'd like to merge this in so we can wrap up review of #193 ASAP

I've responded to/resolved all the outstanding questions as far as I can tell.

static/specification.html
sublimator: A dictionary can NOT be used as an attribute, so whether it's in namespace/scope to be used is a moot point.  
There is a lot of discussion around this in the issues, originally brought up by @sidvishnoi

static/specification.html
sublimator: @adrianhopebailie 

Any thoughts here?

static/specification.html
sublimator: > A payment pointer is a well-defined concept

With a well defined syntax, so it's not exactly 'neat'

static/specification.html
sublimator: @adrianhopebailie 
image

sublimator (changes requested):

static/specification.html
sublimator: Oh, nevermind, in a different view I can see this is for MonetizationEventInit

sublimator: image

Respec compiles just fine

AlexLakatos:
>     2. Introducing `MonetizationCurrencyAmount` or `PaymentCurrencyAmount` and replacing `assetCode` and `assetScale`

The Google team reviewed this, it's good to go. Here's what they said:


>Your plan of introducing a MonetizationCurrencyAmount interface with the same members as PaymentCurrencyAmount and linking to the validation steps in the PaymentRequest API SGTM! 👍 From implementation perspective, you would be able to reuse PaymentRequests' existing validation code, which is great for easier code maintenance.

>The way PaymentRequest API has "solved" the same issue with the total amount on PaymentRequestEvent is by using an object data-type instead. Your solution seems somewhat better. 

>P.S.: Perhaps the PaymentRequest API can consider converting PaymentCurrencyAmount into an interface in the future as well, so that our PaymentRequestEvent is better defined and the PaymentCurrencyAmount would be more easily shared among multiple APIs. No concrete plans for this now, though.

AlexLakatos (approved):

uchibeke (approved):
LGTM

static/specification.html
sublimator: A: Did you get the 5 million dong I sent??
B: Not yet.
A: What?? You said to send ASAP but you haven't even picked it up?
A: WU really gouged me!
A: Was like $295 USD!

What's the 'amountSent' there?

I think "sent" by itself isn't all that clear.

> The "amountSent" in this conversation is 5 million dong, which is a currency used in Vietnam.


no, the amountSent is $295

> Yes, you are correct. In the conversation, the person A mentions that they were charged $295 USD by WU, which implies that the amount sent was indeed $295.

So, is "amountSent" less clear than "sourceAmount" would be here ? 

> In this particular conversation, "amountSent" is clear enough to convey the meaning, as it is referring to the specific amount that was sent via Western Union. However, "sourceAmount" could also be a clear and appropriate alternative.
>
> "SourceAmount" could be used if the conversation involved multiple currencies, and it was necessary to specify the original amount in the currency of the sender. This would be especially useful if the recipient is located in a country with a different currency, as it would allow them to easily convert the amount to their local currency. In this case, "sourceAmount" would be more clear and specific than "amountSent".
>
> Ultimately, the most important thing is to use clear and specific language that accurately conveys the intended meaning.



yes, but you originally said the amountSent was 5 million dong, so how can you argue it's clear enough to convey precisely what was meant ( source or destination asset) 

> You're right, I apologize for the confusion. I misunderstood the context of the conversation you provided earlier. In that specific example, "amountSent" was used to refer to the amount of money that was sent, which was $295 USD.
>
> In general, it's important to use clear and unambiguous language when communicating about financial transactions, especially when dealing with different currencies and assets. Using specific terms like "sourceAmount" and "destinationAmount" can help ensure that both parties understand which asset is being referred to and avoid any potential confusion or misunderstandings.

sublimator:
> For the record, I'm not totally convinced about sourceAmount only, but basically lgtm

Emitting an event with a source amount in a different asset than the page is set up to receive can cause problems because it would require the page to keep a list of exchange rates to do anything meaningful with the amount. For example, if a page is set up to receive payments in USD, but an event is emitted with a source amount in EUR, the page would need to know the exchange rate between USD and EUR to determine the amount of USD received. This can be cumbersome and potentially unreliable.

Even if the source amount can be converted to the desired asset, it cannot be fully verified by the recipient account. The recipient can only really trust and verify what they actually received.
Thus they would need to consult the incomingPayment.

It is possible for a user to send more than one asset as a source amount, which would further complicate the page's ability to use this information. The page would need to know the exchange rate for each asset and then convert them all to the desired asset. It could not simply assume that all amounts will be in a single asset. 

If the method of comparing source amounts requires only the three-letter currency codes, and not the issuer, it can be susceptible to issues. In practice, a USD issued by one exchange may be worth relatively more or less than one issued on another due to various factors such as withdrawal times or limits at the exchange. So, relying solely on the currency code may not be a reliable indicator of the actual value of the asset.

It may be difficult to determine from the monetization event whether an amount is user-initiated or not. One possible way to do this is to set a threshold for the source amount in a particular asset. For example, if the source amount in USD crosses a certain threshold, the page could interpret it as a user-initiated payment and display a "thank you" message. However, this approach may not always be reliable, as there could be other reasons why the source amount crosses the threshold, such as a system error or a test payment.

PR #309, opened 2023-04-11
fix: update docusaurus and fix uphold.md mention

sublimator:


SethFalco (approved):
Sorry for introducing this, and thanks for fixing it!

PR #310, opened 2023-04-11
fix: replace broken uphold link

SethFalco:
Sorry! While removing all Coil related resources, I also removed `uphold.md` as the majority of it's content was regarding Interlegder Payment Pointers which was removed.

However, I missed a reference to `uphold.md`, so there was a broken link which blocked the deploy.

This just changed the link from the internally maintained Uphold docs, to the actual Uphold website.

This should unblock deploying the website.


SethFalco:
Ahh sorry, scrap this!

* Duplicate of https://github.com/WICG/webmonetization/pull/309

Issue #312, opened 2023-06-01
Docusaurus update broke some pages


huijing:
Docusaurus update from 2.0.0-beta-6 to 2.4.0 broke the build (i.e. `npm run build` is broken)
`BrowserOnly` fixed build, but the 3 pages (/exclusive-content, /meta-tag and /prob-revshare) cannot render

PR #313, opened 2023-06-01
Fix import path for Link component

huijing:
Closes https://github.com/WICG/webmonetization/issues/312

This PR resolves the error caused by `Link` component not imported correctly, hence causing the build to fail. Pages render as expected after updating the import path.
`BrowserOnly` is unnecessary and was removed.


AlexLakatos (approved):
LGTM 🍨

PR #331, opened 2023-06-12
Update URLs for link tag generator page

huijing:
We have implemented the redirect from /meta-tag to /link-tag on Cloudflare (please verify by visiting https://webmonetization.org/meta-tag/). This PR is part of the cleanup task and:
- removes meta-tag.js 
- updates the links to /meta-tag to use /link-tag instead


sublimator:
Not using actual redirection ?

The page url just stays on https://webmonetization.org/meta-tag/

sublimator:
href="$YourPaymentPointer" />

Note also that actual payment pointers start with a $
Link tags accept only well formed urls
It might pay to have the initial examples/placeholders correct

sublimator (approved):

huijing:
> Not using actual redirection ? The page url just stays on https://webmonetization.org/meta-tag/

Ah, made a typo on the redirect rule up on Cloudflare. It's corrected now.

sublimator:
Just testing if anyone awake :) ?

PR #346, opened 2023-07-05
Add note for allow attribute

### LINKED ISSUE #244: allow attribute and fetch/process
In the list under "The appropriate times to fetch and process this type of link is: "

Shouldn't we also mention the allow attribute of an iframe being changed? I mean, if allow="monetization" is removed the monetization stream should end, right?

huijing:
Closes #244 

This PR adds a note that explains the adding or removing the monetization attribute has no effect on an already-loaded document.

1. Pull down this branch
2. Navigate to http://localhost:3000/specification.html#permissions-policy and check that the note is rendering correctly (or you could probably use the deploy preview since we turned that on ¯\\\_(ツ)_/¯)

### DIFF
--- a/static/specification/index.html
+++ b/static/specification/index.html
@@ -1,1853 +1,1856 @@
 <!DOCTYPE html><html lang="en"><head>
-<meta charset="utf-8">
-<meta name="generator" content="ReSpec 34.1.4">
-<meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
-<style>
-span.example-title{text-transform:none}
-:is(aside,div).example,div.illegal-example{padding:.5em;margin:1em 0;position:relative;clear:both}
-div.illegal-example{color:red}
-div.illegal-example p{color:#000}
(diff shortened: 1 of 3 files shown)


static/specification/index.html
```diff
@@ -1,1853 +1,1856 @@
...
-    </p>
-    <dl>
-      
-      <dt>Latest published version:</dt><dd>
-              <a href="https://www.w3.org/web-monetization/">https://www.w3.org/web-monetization/</a>
-            </dd>
-      <dt>Latest editor's draft:</dt><dd><a href="https://webmonetization.org/specification">https://webmonetization.org/specification</a></dd>
-      
-      
+  <meta charset="utf-8">
```
huijing, line 2: unfortunately, this file got prettified 🙇🏻‍♀️

sublimator:
Prettier them all during a lull?

AlexLakatos:
I think we need to add the same note in the docs as well.

AlexLakatos (approved):
LGTM 📓

PR #347, opened 2023-07-05
Add libraries page to the docs

### LINKED ISSUE #263: [Docs] Web Monetization API - Libraries
- [x] Section for Libraries on main page
    - Issue with current main page is that the libraries are all relevant for v1.0 only
    - We will have to keep this section like it is until there are v2.0 libraries available
- [x] Separate page with links to different libraries 
    - For now, we will duplicate the home page tools section under Resources but label them as v1.0

huijing:
Closes #263 

This PR adds a page to the documentation's Resource section for libraries. It lists the libraries in a table form, with fields that show which version of the specification the libraries are compatible with and a repository link.

1. Pull down this branch
2. Navigate to http://localhost:3000/docs/libraries and check that the page is rendering correctly (or you could probably use the deploy preview since we turned that on ¯\\\_(ツ)_/¯)


src/components/Libraries.js
```diff
@@ -0,0 +1,44 @@
...
+    return data.map(({ name, desc, link, version }) => {
+      return (
+        <tr key={name}>
+          <td>
+            <strong>{name}</strong>
+            <br />
+            {desc}
+          </td>
+          <td class='cell-center'>{version.includes(2) ? '✅' : '❌'}</td>
+          <td class='cell-center'>{version.includes(1) ? '✅' : '❌'}</td>
```
AlexLakatos, line 17: This breaks down if we ever move to 2.1 or something like that. Can we make it a bit more robust? Maybe check for the dot as well, and see if the version starts with, not only includes the number?

AlexLakatos (changes requested):

src/components/Libraries.js
huijing: What's our versioning strategy?
Because this table might start looking ridiculous after 5 versions.

AlexLakatos: It all depends on how massive the change recommendations are after the browser implementation we're experimenting with. The first time we did this, we've had to move from v1 to v2. If the changer aren't big, we'll probably do v2.1, and that's going to be the only "2" version we'll talk about. So the table might be "v2.1, v1" without mentioning v2.0

AlexLakatos (approved):
LGTM 🦺

PR #349, opened 2023-07-05
Update folder structure for specification

### LINKED ISSUE #222: Deploy snapshot of spec, not raw ReSpec doc
It's nicer for users see see a built version of the spec, rather than the raw ReSpec version. 

Instructions (including CLI) for creating a snapshot are here:
https://respec.org/docs/#static-snapshots

Alternatively, we can use:
https://github.com/w3c/spec-prod

To generate the static snapshot.

### LINKED ISSUE #339: Introduce versioning for specification.html
If we reference the way the CSSWG does their specifications, you will see that every specification has links to previous versions of the specification (for example: https://www.w3.org/TR/selectors-4/). This is a good practice we would also like to adopt.
For starters, we need to make the previous version of the specification available again under a "Previous Versions" section in the top of the document.

Related to #222 
Essentially, we should move the ReSpec version out of the static folder and continue to do specification work on it, while the version(s) in the static should be the ones exported from the ReSpec version.

huijing:
Closes #222 
Closes #339

This PR renames the original specification.html file into specification-respec.html and replaces public links to the specification to the statically generated version instead. This also introduces a folder structure that allows the file to be accessed without the `.html` at the end of the URL.
The previous version of the specification was also generated for record-keeping purposes. And the instructions for developing the specification locally have been updated.

1. Pull down this branch
2. Navigate to http://localhost:3000/specification and check that the specification is rendering correctly and it is NOT the respec version
3. Navigate to http://localhost:3000/specification-respec.html and check that it is the ReSpec version. 
4. Navigate to https://localhost:3000/specification/versions/CG-DRAFT-web-monetization-20210317.html and check that it is version 1.0 of the specification 

(or you could probably use the deploy preview since we turned that on ¯\\\_(ツ)_/¯)


README.md
```diff
@@ -28,6 +28,16 @@ $ yarn
 $ yarn start
 ```
 
+## Specification Development
```
huijing, line 31: @AlexLakatos this is explanation for the changes.

AlexLakatos (approved):
LGTM 🏬

PR #355, opened 2023-07-11
Resolve github actions failure

huijing:
After messing around in a test repo, I finally figured out the correct YML steps for the thing to run correctly (see https://github.com/huijing/octokit-test/actions)

### DIFF
--- a/.github/actions/get-contributors.js
+++ b/.github/actions/get-contributors.js
@@ -15,10 +15,7 @@ async function getContributors() {
     }
   )
   const contributors = JSON.stringify(data)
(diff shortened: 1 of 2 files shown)


.github/actions/get-contributors.js
```diff
@@ -15,10 +15,7 @@ async function getContributors() {
     }
   )
   const contributors = JSON.stringify(data)
-  fs.writeFileSync(
```
huijing, line 18: apparently, github actions runs from root.

AlexLakatos (approved):
LGTM 🪗

PR #363, opened 2023-07-18
Add gh action for deployment

### LINKED ISSUE #360: Replace circle-ci deployment with GitHub Actions
Circle CI is running out of memory, and failing to deploy the builds.

We should take this opportunity to swap deployments to GitHub actions.

The current setup uses the `yarn deploy` command (build the website, and commit it to the gh-pages branch, which takes care of the hosting). I think we should be able to replicate that for GitHub Actions.

huijing:
Closes #360


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,40 @@
...
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 18
+          cache: yarn
+
+      - name: Install dependencies
+        run: yarn install
+      - name: Build website
+        run: yarn build
```
AlexLakatos, line 24: I think just doing yarn deploy here would do the trick, without having to use the peaceiris/actions-gh-pages@v3 step

PR #395, opened 2023-09-19
Test mermaid on plain md

huijing:
Just testing Netlify


huijing:
Markdown was not the problem.

PR #399, opened 2023-10-09
chore: refactor custom header implementation

### LINKED ISSUE #398: Refactor custom header after Starlight minor version bump
The latest version of Starlight (https://github.com/withastro/starlight/releases/tag/%40astrojs%2Fstarlight%400.11.0) released its long-awaited feature for customising Starlight components. Thus far, we have been using a community-sourced "hack" method of doing this (see https://gist.github.com/giuseppelt/7f918a3ac02a011d76811ae472f8bf09), with the intention of migrating to the "official" method of component overriding the moment it was baked. The time has come for this to happen.
We will also need to make some styling tweaks and checks because the API of how the Starlight internals (i.e. CSS classes, HTML structure) would have changed a bit due to the additional feature.

huijing:
Closes #398 

This PR removes the previous custom header implementation files, and bumps dependencies to the latest version.


astro.config.mjs
```diff
@@ -212,12 +213,6 @@ export default defineConfig({
     react(),
     astroI18next(),
   ],
-  // Process images with sharp: https://docs.astro.build/en/guides/assets/#using-sharp
```
huijing, line 215: We are not running our images through the sharp pipeline.

src/components/docs/Header.astro
```diff
@@ -1,22 +1,19 @@
 ---
+import type { Props } from '@astrojs/starlight/props';
```
huijing, line 2: This is the official method of doing Starlight component override, see https://starlight.astro.build/guides/overriding-components/ for documentation.

sabineschaller (approved):

PR #406, opened 2023-10-28
CSS changes

codeNaitik:
Done CSS change in webMonetization repo, include some good quality and optimize CSS.


huijing:
Thank you for your contribution, however, this change does not appear to be beneficial. If the proposal is to use transforms instead of positioning for animation then yes, but in this case, there is no animation going on.

PR #419, opened 2023-12-14
chore: updated libraries with latest support info available

devcer:
* The owner has transferred ownership for eleventy plugin to me -> https://github.com/devcer/eleventy-plugin-monetization So updated the url and the support for v2
* Astro plugin is also published. Please let me know if I do have to change the order in the list.


huijing (approved):
Rerun deployment on Netlify after clearing cache, and deploy preview was successful. Reviewed deployment preview, checked home page and libraries table. Both render as expected.

PR #423, opened 2024-01-11
fix: use proper feature detection in is-supported-example

jeremiahlee:
Change wallet address to match other examples, removing the .js prefix which could be confusing.

Remove deprecated methods for detecting support of a feature in a user agent.

Remove DOM query in favor of faster temporary element creation.

Tangential issues:
- Fixed local dev issues related to updating the spec now that Astro is used
- Updated readme instructions to use the spec build script
- Removed Prettier from spec build script, as it conflicts with W3C style for examples


src/pages/specification/includes/is-supported-example.html
```diff
@@ -1,20 +1,8 @@
-<link rel="monetization" href="https://example.com/monetization.js" />
+<link rel="monetization" href="https://example.com/pay">
 <script>
-  // Checking via DOMTokenList
-  const link =
-    document.querySelector('link[rel="monetization"]') ||
-    document.createElement('link')
-  if (link.relList.supports('monetization')) {
```
sublimator, line 7: This example always bugged me - 3 lines logged

src/pages/specification/specification-respec.html
```diff
@@ -117,12 +117,9 @@ <h3>
           Checking if Web Monetization is supported
         </h3>
         <p>
-          There are multiple ways to check if Web Monetization is supported.
```
sublimator, line 120: iirc ?? this file is generated, so you'll want to edit the base file, and then run the respec.sh command

sublimator: nevermind, this /IS/ the source
https://github.com/WICG/webmonetization/blob/main/src/pages/specification/respec.sh#L21

sublimator (approved):

PR #425, opened 2024-01-11
fix: remove trailing slashes in examples

jeremiahlee:
Trailing slashes are advised against in HTML Standard for void elements and using them causes info notices in the W3C validator. To provide the most correct examples, we should not use trailing slashes.

https://github.com/validator/validator/wiki/Markup-%C2%BB-Void-elements


src/pages/specification/includes/media-example.html
```diff
@@ -1,11 +1,11 @@
 <video src="myvideo" onmonetization="sayThanks(this)">
-  <link rel="monetization" href="https://example.com/video/pay" />
+  <link rel="monetization" href="https://example.com/video/pay">
 </video>
```
sublimator, line 4: One thing to check is the tidy configuration

sublimator: https://github.com/WICG/webmonetization/blob/main/src/pages/specification/tidyconfig.txt

jeremiahlee:
Currently blocked. This requires regenerating the ReSpec output, but this should be done after #423 is merged and then this branch updated.

sublimator (approved):

PR #426, opened 2024-01-11
docs: updates to readme

melissahenderson:


melissahenderson:
@huijing Do you mind giving this a look over when you're feeling better? I removed the "Web Monetization Specification" section and would like your opinion on whether that was a good idea. If not, I can add it back.

huijing:
I think it's fine to remove, but could we keep those details in a separate MD file maybe?

melissahenderson:
This merge will include fixes for #431

huijing (approved):
LGTM

PR #451, opened 2024-04-30
Update README.md to include catchup call details

sabineschaller:
Added info on call plus link to add to calendar.


huijing (approved):
LGTM

PR #458, opened 2024-06-18
fix(spec): make bubbling clearer for `onmonetization`

sidvishnoi:
Also removed duplicate content in head.

### DIFF
--- a/src/pages/specification/specification-respec.html
+++ b/src/pages/specification/specification-respec.html
@@ -8,12 +8,6 @@
     </title>
     <script src="https://www.w3.org/Tools/respec/respec-w3c" class="remove"
     defer></script>
-    <meta charset="utf-8">
-    <title>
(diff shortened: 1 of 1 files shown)


sublimator:
Won't somebody think of the grandchildren?

sidvishnoi:
Was adding `onmonetization` to `<body>` while the `<link>` was in `<head>` (and I wasn't alone doing that),

sidvishnoi:
Oh I need to generate specification manually? It's not in deploy preview.

sublimator:
Won't somebody think of the respec job?

sidvishnoi:
I don't like adding that file to commit log. I'll do the CI first.

BGAK2024 (approved):

sublimator:
Once upon a time

Issue #401, opened 2023-10-13
Translate "Intro to Web Monetization" page


huijing:
We would like this page: https://github.com/WICG/webmonetization/blob/main/src/content/docs/docs.mdx translated into the following languages:
- [ ] Spanish
- [ ] Portuguese
- [ ] Arabic
- [ ] Japanese
- [ ] Chinese

Our contribution guide for this specific task can be found at: https://github.com/WICG/webmonetization#adding-translated-content

If you are willing and able to contribute translations in any other language than those listed above, we welcome your contribution with open arms and much gratitude as well. 🙇🏻‍♀️

IrkenStudio:
Hi, Can you assign this issue to me, I can do the Spanish translation if it hasnt been done yet. I am a native spanish speaker and billingual proficiency in english.

Urva2:
I want to fix this issue.

Smitg2:
i am intrested in doing this. please give me this task ....

huijing:
> Hi, Can you assign this issue to me, I can do the Spanish translation if it hasnt been done yet. I am a native spanish speaker and billingual proficiency in english.

Hello @IrkenStudio :wave:
Thank you so much for volunteering for this. I've assigned the issue to you. Please let me know if you run into any issues with the local setup, or if the README instructions are not clear enough.

huijing:
@Urva2 and @Smitg2 

May I know which language you would like to do the translations for?

melissahenderson:
It looks like someone has a PR for the Spanish version, but all the checks failed. https://github.com/WICG/webmonetization/pull/407

huijing:
> It looks like someone has a PR for the Spanish version, but all the checks failed. #407

Do we have someone who can proof-read the Spanish? If so, I probably can fix the PR to pass.

melissahenderson:
@AlexLakatos See above please

AlexLakatos:
Nestor has done the interledger.org site, we can ask him.

melissahenderson:
Spanish translation is complete

alancpazetto:
Hi! I've discovering this project recently and I'm interested to help.

Can you assign portugues translate to me? In case, Brazilian Portuguese.

Note, second link in issue description is broken: https://github.com/WICG/webmonetization#adding-translated-content

alancpazetto:
Hi! 

PR openned with PT-BR translation: #474

melissahenderson:
[...]

PR #474, opened 2024-08-22
docs: added pt-br index page

alancpazetto:
This PR adds docs index page in Portuguese Brazil.

It's part of #401


melissahenderson:
The "Intro to Web Monetization" page no longer exists in this translated form. We also removed locales functionality from the site at some point. Closing PR.

PR #485, opened 2024-12-13
chore: basic dependency bump

JoblersTune:
There have been a few difficulties with this dependency bump.

Firstly, Starlight, as of now, has a compatibility issue with astro@5. This is being addressed (see [https://github.com/withastro/starlight/pull/2612](https://github.com/withastro/starlight/pull/2612))

Secondly, bumping up the @astrojs/starlight dependency caused compatibility issues with the `HeadHrefLangs` component that is part of the astro-i18next module. Unfortunately, it seems this module is no longer being maintained. As such, I've removed it for now. HeadHrefLangs is a component from astro-i18next that generated <link> tags with hreflang attributes for SEO, specifying alternate language versions of a webpage. This means we need to investigate a new way to build this in and probably use either a different community based plugin for this, or move over to Astro's Built-in i18n Support. This also has limitations which is why it needs more work to figure out the best path.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,21 +8,21 @@
     "astro": "astro"
   },
   "dependencies": {
-    "@astrojs/react": "^3.6.2",
-    "@astrojs/starlight": "^0.26.1",
-    "@interledger/docs-design-system": "^0.5.2",
-    "@types/react": "^18.3.3",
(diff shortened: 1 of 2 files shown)


huijing (approved):
LGTM

Issue #491, opened 2025-01-13
[Playground]: provide wallet address via URL parameter


sidvishnoi:
**Is your feature request related to a problem? Please describe.**
When running tests for Web Monetization, we use playground and add link tags via the form. This is good when we're testing dynamic link tag addition, but we would also like to test cases when the link tag is already added on page when it loads.
This would also allow us embedding playground as iframe for iframe tests.

**Describe the solution you'd like**
Ability to add wallet address via URL parameter. Multiple instances of query parameter should add multiple wallet addresses.

**Describe alternatives you've considered**
Host our own pages for end-to-end tests in extension somewhere. But we want them hosted on https:// URLs, as that's where WM is supported.

**Additional context**
Question: is the website rendered via SSR or is it CSR/prerendered only?
Note: I can send PR.

sidvishnoi:
Or, I can create a worker, which returns basic HTML for above cases?

JoblersTune:
Astro generates static content by default. However, it uses an island architecture that allows client-side components to be hydrated after the page loads. Given this architecture, adding <link> tags dynamically based on URL parameters should work. 

PRs are welcomed! And if I'm needed for calls, context, work or reviews please tag me or ping me.

JoblersTune:
SSR is still possible with some extra configuration, see [https://docs.astro.build/en/guides/on-demand-rendering/](https://docs.astro.build/en/guides/on-demand-rendering/)

And when a component needs client side rendering you include the client:load directive when you use the component, see [https://docs.astro.build/en/reference/directives-reference/#clientload](https://docs.astro.build/en/reference/directives-reference/#clientload) and an example in our site [https://github.com/WICG/webmonetization/blob/main/src/pages/prob-revshare.astro#L21](https://github.com/WICG/webmonetization/blob/main/src/pages/prob-revshare.astro#L21)

PR #492, opened 2025-01-14
feat(/play): add wallet addresses from URL parameters

sidvishnoi:
Closes https://github.com/WICG/webmonetization/issues/491

Add `?wa=walletAddressUrl` (multiple ok) to add link tags on page load. Matches validations like form (except must be URL, not payment pointers). If WM is not supported, tags are not added.
[Example](https://deploy-preview-492--webmonetization-preview.netlify.app/play/?wa=https://ilp.interledger-test.dev/sid&wa=https://ilp.interledger-test.dev/sid-mxn-1&wa=https://ilp.interledger-test.dev/sid)

### DIFF
--- a/src/pages/play.astro
+++ b/src/pages/play.astro
@@ -210,6 +210,7 @@ const description = t("site.description");
   if (window.MonetizationEvent) {
     submitButton.disabled = false
     notice.style.display = 'none'
+    addTagsFromURL()
   } else {
     setTimeout(() => {
       if (!window.MonetizationEvent) {
@@ -220,6 +221,7 @@ const description = t("site.description");
       } else {
         submitButton.disabled = false
         notice.style.display = 'none'
+        addTagsFromURL()
       }
     }, 500)
   }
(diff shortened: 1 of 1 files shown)


JoblersTune:
Does issue #491 still need to remain open to cover testing when <link> tags are added on page load?

JoblersTune:
Since you have it listed as a BUG that you've restricted the same link tag being added twice I thought I'd bring it up that if the tag is in the URL params and then added by the form it is added twice.
image

sidvishnoi:
I think can close #491 with this.

On adding same URL with form, right now it'll show "already added" validation error. I think playground shouldn't artificially restrict what can go on the page (as user can duplicate link tags in HTML too, and WM won't complain) - that's why I've added that comment as a bug. Maybe a warning that it's already added would be fine, but shouldn't restrict from adding. I'll file follow-up issue and send PR for that.

JoblersTune (approved):

PR #504, opened 2025-03-12
feat: new tools page that loads WM tools

lengyel-arpad85:


huijing (approved):
Screenshot 2025-03-12 at 7 10 48 PM

1. Some of the buttons/fields have transparent background, can we standardise them to be white?
2. The dropdown for choosing colour for background and text has a scrollbar but this problem only shows up on Chrome, specifically tested using Version 136.0.7063.0 (Official Build) canary (arm64)

All feedback is non-blocking, so approving this PR.

src/pages/tools.astro
```diff
@@ -0,0 +1,39 @@
...
+      src="https://wmtools.interledger-test.dev?contentOnly"
+      title="Publisher tools"
+      loading="eager"
+    />
+  </section>
+</Base>
+
+<style>
+  body {
+    background-image: url('/img/bg-tile.svg');
```
sidvishnoi, line 23: Should either remove this background, or add a `#ffffff` on iframe, otherwise there's 2x  2x web/mash background.

src/pages/tools.astro
```diff
@@ -0,0 +1,39 @@
...
+    background-size: 25em;
+  }
+
+  .iframe_content {
+    width: 100%;
+    height: calc(100vh - 190px);
```
sidvishnoi, line 29: Need to be careful with this on mobile.

PR #498, opened 2025-01-28
docs: reorganize and revise user docs

melissahenderson:
This PR replaces #473. The structure was approved by Bibi on 24 Jan 2025.

There are still a few pages that need content, so this PR is not ready for merging yet.

PR #509, opened 2025-03-18
chore: move all tools under /tools

AlexLakatos:
### Changes requested

- Move rev-share and linktag generators under /tools

### DIFF
--- a/astro.config.mjs
+++ b/astro.config.mjs
@@ -235,7 +235,7 @@ export default defineConfig({
   ],
   redirects: {
     "/link-tag": "/tools/link-tag",
-    "/revshare": "/tools/prob-revshare",
+    "/prob-revshare": "/tools/prob-revshare",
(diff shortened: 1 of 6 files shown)


astro.config.mjs
```diff
@@ -234,6 +234,8 @@ export default defineConfig({
     astroI18next(),
   ],
   redirects: {
+    "/link-tag": "/tools/link-tag",
+    "/revshare": "/tools/prob-revshare",
```
sidvishnoi, line 238: `/revshare` doesn't exist.

```suggestion
    "/prob-revshare": "/tools/prob-revshare",
```

sidvishnoi (approved):
There's a reference to `/prob-revshare`  in `src/content/docs/docs/guides/set-up-probabilistic-revenue-sharing.mdx`, but let's not do that now, as there will be a in merge conflict in https://github.com/WICG/webmonetization/pull/498 otherwise.

DarianM (approved):

PR #515, opened 2025-04-02
Fix a technically incorrect sentence

sensingturtle:


JoblersTune:
Thanks @sensingturtle, good catch.

PR #530, opened 2025-05-19
Styles and payment pointer references updates

melissahenderson:

PR #531, opened 2025-05-20
feat: added a component used for text where emphasis is needed

JoblersTune:
For the content owner and content supporter overview pages it would be nice to have text that brings emphasis and clearly separates points but this text is not strictly a heading so I made a local component that can style the text as if it were a heading but without the stylish underline.

Usage:
```
import EmphasisText from '/src/components/docs/EmphasisText.astro'

<EmphasisText text='The network of Web Monetization-enabled wallet providers is nascent, but growing.' />
```

### DIFF
--- a/src/components/docs/EmphasisText.astro
+++ b/src/components/docs/EmphasisText.astro
@@ -1,0 +1,12 @@
+---
+const { text } = Astro.props;
+---
+<p>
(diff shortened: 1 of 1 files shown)


melissahenderson:
Added this in with #530

PR #548, opened 2025-07-15
Patch CSP's "Get directive fallback list" to include `monetization-src`

### LINKED ISSUE #536: Define directive fallback list for `monetization-src` CSP directive
In CSP there's a [directive fallback list ](https://w3c.github.io/webappsec-csp/#directive-fallback-list) defined for various CSP directives. This spec should probably define one for   the `monetization-src` directive.

Something like:

Return << "monetization-src", "default-src" >>.

If it's not desirable to fallback to default-src I think this should be called out explicitly with a strong reasoning.

lukewarlow:
Fixes #536

### DIFF
--- a/specification/index.html
+++ b/specification/index.html
@@ -852,6 +852,31 @@
             </li>
           </ol>
         </section>
+        <section data-cite="CSP">
+          <h4>
+            Get fetch directive fallback list
+          </h4>
+          <p>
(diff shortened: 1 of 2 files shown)


lukewarlow:
cc @sidvishnoi

MichaelKohler (approved):

PR #569, opened 2025-08-19
chore: clean up

JoblersTune:


workers/exclusive-content/README.md
sidvishnoi, line 1: We don't have a relevant tool for this in publisher tools repo yet. Are we planning to deprecate this?

sidvishnoi (approved):

workers/exclusive-content/README.md
JoblersTune: I'd argue it doesn't belong here any more. It will remain in git history. But I'm happy to leave it in if required. What do you think? Or should we ask anyone else?

JoblersTune: Does it provide current functionality that people might be using?

sidvishnoi: I see whatever functionality it supported, relied on receipt-verifier service, and its source code is not here anymore.

Let me tag @DarianM here so he knows when and how this source was removed if we want it in tools repo in future, and we can remove it from here now.

JoblersTune: I can leave a comment in my commit message of the redesign to mark it easily searchable. Like, 

"Removed all tools functionality
Removed exclusive content code"

DarianM: This cleanup contains `probabilistic-revshare` worker code that will eventually need to be migrated to tools.

Until that happens, the CLF worker will run normally and will not affect the revenue sharing tool.

Good notification 👍 now it will show me how to find the things I need more easily.

PR #586, opened 2025-10-01
chore: Improve carousel navigation

Anca2022:
- Turn carousel dots into buttons and allow navigation on click


Anca2022:
Opening new [PR] (https://github.com/WICG/webmonetization/pull/591)

PR #592, opened 2025-10-06
fix(/play): apply buttons redesign updates

Anca2022:
Linear issue: [ENG-587](https://linear.app/interledger/issue/ENG-587/wm-redesign-apply-style-updates-to-playground-buttons) & [WM-264](https://linear.app/interledger/issue/WM-264/wm-redesign-horizontal-overflow-on-play-page-on-small-screens)

- add SVG icons to Playground buttons
- turn Enabled/Disabled Playground button into a toggle
- adjust .link-events grid dimensions to eliminate horizontal overflow


src/pages/play.astro
```diff
@@ -222,19 +258,28 @@ const t = useTranslations(lang)
     const footer = document.createElement('div')
     footer.classList.add('log-footer')
 
-    const disableButton = document.createElement('button')
-    disableButton.textContent = 'Disable'
-    disableButton.classList.add('btn', 'tools-secondary-btn')
-    disableButton.addEventListener('click', () => {
+    const switchButton = document.createElement('div')
+    switchButton.innerHTML = `
+      <label class="wallet-switch">
```
sidvishnoi: Nit: let's create `const switchButton = document.createElement('label')` and skip the extra `div`?

src/pages/play.astro
```diff
@@ -222,19 +258,28 @@ const t = useTranslations(lang)
...
       ) as HTMLLinkElement
       link.disabled = !link.disabled
-      disableButton.textContent = link.disabled ? 'Enable' : 'Disable'
     })
 
     const removeButton = document.createElement('button')
-    removeButton.textContent = 'Remove'
+    removeButton.innerHTML = `<svg aria-hidden="true" width="20" height="20" viewBox="0 0 24 24" fill="none">
+      <path d="M7.616 20C7.168 20 6.78667 19.8426 6.472 19.528C6.15733 19.2133 6 18.8323 6 18.385V5.99998H5V4.99998H9V4.22998H15V4.99998H19V5.99998H18V18.385C18 18.845 17.846 19.2293 17.538 19.538C17.23 19.8466 16.8453 20.0006 16.384 20H7.616ZM17 5.99998H7V18.385C7 18.5643 7.05767 18.7116 7.173 18.827C7.28833 18.9423 7.436 19 7.616 19H16.385C16.5383 19 16.6793 18.936 16.808 18.808C16.9367 18.68 17.0007 18.5386 17 18.384V5.99998ZM9.808 17H10.808V7.99998H9.808V17ZM13.192 17H14.192V7.99998H13.192V17Z" fill="currentColor"/>
+      </svg>
```
sidvishnoi: Can you extract these SVGs into a top-level constant? Or even better, add them to `src/assets` and then we can either do an `?inline` import of use `` to use them.

src/pages/play.astro
```diff
@@ -63,7 +63,7 @@ const t = useTranslations(lang)
   }
   .link-events {
     display: grid;
-    grid-template-columns: repeat(auto-fit, minmax(30rem, 1fr));
+    grid-template-columns: repeat(auto-fit, minmax(16.5rem, 1fr));
```
sidvishnoi: This might be too small. We want some extra width for each item, so we can comfortably read the logs, and also not overflow on wallet address.

Image
At this particular screen size (13" Apple M1 Air), I'd expect to see only 2 items, not 4 in single row.

sidvishnoi (changes requested):

src/pages/play.astro
Anca2022: I'm changing back to 30rem by default and setting min to 16.5rem on smaller screens (where the min of 30rem is causing a horizontal overflow)

PR #594, opened 2025-10-07
chore: update dropdown menu images

Anca2022:
Linear issue:[ ENG-584](https://linear.app/interledger/issue/ENG-584/wm-redesign-update-drop-down-menu-images)


src/components/pages/NavToolsDropdown.astro
```diff
@@ -20,13 +20,22 @@ const isActive = (path: string): boolean =>
...
+                  src={menuItem.imageInactive.src}
+                />
+                <img
+                  width={menuItem.imageInactive.width}
+                  height={menuItem.imageInactive.height}
+                  class='tool-icon hovered-image'
+                  alt=''
+                  src={menuItem.imageActive.src}
+                />
+              </div>
```
sidvishnoi, line 38: Can we lazy load or reduce fetchpriority of these images, and not require inlining them (at least for hover ones)? I think these images are very much cacheable, and if they can load before user opens the nav some way, I think that can be worthwhile (ok as a follow up).

src/assets/dropdown-supporter-inactive.png
sidvishnoi, line 1: I'm not finding a lot of difference between these two images. Can you tell me how they're different?

src/components/pages/NavToolsDropdown.astro
```diff
@@ -20,13 +20,22 @@ const isActive = (path: string): boolean =>
...
           >
             <>
-              <img
-                width={menuItem.image.width}
-                height={menuItem.image.height}
-                class='tool-icon'
-                alt=''
-                src={menuItem.image.src}
-              />
+              <div class='tool-icon-container'>
```
sidvishnoi, line 23: Can we avoid this `div` easily?

src/data/nav.ts
```diff
@@ -1,52 +1,70 @@
...
+    src: string
+    width: number
+    height: number
+  }
+  imageActive: {
+    src: string
+  }
+  translationKey: ToolLabelKey
+  path: string
+  event: string
```
sidvishnoi: Is this some tabs vs spaces issue? Can we avoid changing whitespace in this PR, and make it consistent across all files in a single separate PR?

sidvishnoi (commented):
Looking good. Left few minor suggestions.

src/data/nav.ts
Anca2022: It's prettier - "semi: false, tabWidth: 2"

src/assets/dropdown-supporter-inactive.png
Anca2022: The blurred margin is white on one, gray on the other - it matches the colors of the menu on static / active state

Anca2022: It would look this way if we used the same image for hover: 
image

sidvishnoi: @DarianM FYI ^

DarianM: yes, we need update on our end too

src/data/nav.ts
```diff
@@ -1,16 +1,21 @@
...
 interface ToolsSubmenuItem {
-    image: {
+    imageInactive: {
         src: string,
         width: number,
         height: number
     },
+    imageActive: {
+        src: string
+    },
```
sidvishnoi, line 18: Consider using: `image` and `imageInteractive` (or something).
Active/Inactive can be confusing.

src/components/pages/NavToolsDropdown.astro
```diff
@@ -20,13 +20,24 @@ const isActive = (path: string): boolean =>
...
+                <img
+                  width={menuItem.image.width}
+                  height={menuItem.image.height}
+                  class='tool-icon default-image'
+                  alt=''
+                  src={menuItem.image.src}
+                  loading='lazy'
```
sidvishnoi: When using inline (base64 `data:` URIs), lazy won't have any benefit, as image is already loaded.
```suggestion
```

sidvishnoi (approved):
LGTM. But let's merge it along with tools, so the website overall has same look and feel for nav.

PR #595, opened 2025-10-09
feat(FAQ): create FAQ page for WM website

Anca2022:
Linear issue[ WM-254 WM Redesign - FAQ page](https://linear.app/interledger/issue/WM-254/wm-redesign-faq-page)
Figma file [here](https://www.figma.com/design/0S8Oj8ZB5MvkGtIrVRswDH/Web-Monetization?node-id=5978-134473&t=ZoqXwbOgMX79KCNj-1)

-  implement page layout and styling as per Figma design
- accordion accessibility
- added real FAQ data to the page

### DIFF
--- a/src/components/pages/Footer.astro
+++ b/src/components/pages/Footer.astro
@@ -39,6 +39,14 @@ const t = useTranslations(lang)
             target='_blank'>{t('footer.menu.blog')}</a
           >
         </li>
+        <li>
+          <a
+            href='/faq'
+            data-umami-event='Footer - FAQ page'
+            rel='noopener noreferrer'
+            target='_blank'>{t('footer.menu.faq')}</a
+          >
+        </li>
       </ul>
       <ul>
         <li>
--- a/src/i18n/ui.ts
+++ b/src/i18n/ui.ts
@@ -38,6 +38,7 @@ export const ui = {
     'footer.menu.community': 'Community',
     'footer.menu.github': 'GitHub',
     'footer.menu.blog': 'Blog',
+    'footer.menu.faq': 'FAQ',
     'footer.menu.summit': 'Summit',
     'footer.menu.privacy_policy': 'Privacy Policy',
     'footer.menu.cookie_policy': 'Cookie Policy',


src/pages/faq.astro
```diff
@@ -0,0 +1,213 @@
...
+                <div
+                  class='faq-answer'
+                  id={`faq-answer-${category.id}-${i}`}
+                  aria-hidden='true'
+                >
+                  {item.answer.map((answer) => (
+                    <p set:html={answer} />
+                  ))}
+                </div>
+              </div>
```
sidvishnoi: Can we use `<details>` element for this?

src/pages/faq.astro
```diff
@@ -0,0 +1,213 @@
...
+    background-color: var(--button-secondary-color);
+    transition: transform 0.2s ease-in-out;
+  }
+
+  .faq-item button.expanded {
+    font: var(--font-standard-bold);
+  }
+
+  .faq-item button.expanded::after {
+    transform: rotate(0deg);
```
sidvishnoi: As per design in Figma, we by default show ➕ and once expanded, show ❌ . Right now, this appears to be flipped.

src/pages/faq.astro
```diff
@@ -0,0 +1,213 @@
...
+    background-color: white;
+    margin-bottom: var(--spacing-xs);
+    padding: var(--padding-md);
+  }
+
+  .faq-item:last-of-type {
+    margin-bottom: 0;
+  }
+
+  .faq-item h4 {
```
sidvishnoi: These need vertical centering when in non-expanded state.

src/data/faq.ts
sidvishnoi, line 1: Given we include enough markup in each answer, would it be better if we create a component `<FaqItem title="title">content here</FaqItem>` instead of using TypeScript data?

Data approach can help us in future though - like we can source data from a CMS.

Trade-offs!

Let's keep the data approach for now then. Keep this comment open so future-us know we considered this.

src/data/faq.ts
```diff
@@ -0,0 +1,151 @@
...
+  answer: string[]
+}
+
+interface FaqCategory {
+  id: number
+  title: string
+  questions: FaqQuestion[]
+}
+
+const faqContent: FaqCategory[] = [
```
sidvishnoi, line 146: Nit: Consider moving each "section" into a separate `const` and then create the `faqContent` list of out it. We can avoid one level of nesting in content array this way, and make data easier to follow.

PR #598, opened 2025-10-14
fix(/faq): expand clickable area on <summary> to include "+" icon

Anca2022:
Linear issue [WM-267](https://linear.app/interledger/issue/WM-267/faq-page-open-details)
Follow up of https://github.com/WICG/webmonetization/pull/595.

### DIFF
--- a/src/pages/faq.astro
+++ b/src/pages/faq.astro
@@ -172,8 +172,8 @@ const t = useTranslations(lang)
     }
 
     details.faq-item summary {
-      padding: var(--padding-2xl) calc(2 * var(--padding-2xl)) var(--padding-2xl)
-        var(--padding-2xl);
+      padding: var(--padding-2xl);
+      padding-right: calc(2 * var(--padding-2xl));
     }
 
     details.faq-item[open] summary {


src/pages/faq.astro
```diff
@@ -175,18 +170,28 @@ const t = useTranslations(lang)
       font: var(--font-h2);
     }
 
-    details.faq-item {
-      padding: var(--padding-2xl);
+    details.faq-item summary {
+      padding: var(--padding-2xl) calc(2 * var(--padding-2xl)) var(--padding-md)
```
sidvishnoi: We can add more to bottom padding here (I think 2xl on bottom as well) and remove it from `<details>`. Then we can click anywhere in non-expanded state to open it.

src/pages/faq.astro
```diff
@@ -175,18 +170,28 @@ const t = useTranslations(lang)
...
     }
+
+    details.faq-item p {
+      padding: 0 calc(2 * var(--padding-2xl)) var(--padding-md)
+        var(--padding-2xl);
+    }
+
+    details.faq-item p:last-of-type {
+      padding-bottom: calc(var(--padding-2xl) + var(--padding-xs));
+    }
```
sidvishnoi: Instead of target `p`, we can add padding once to `details > div` or (`details > div > div` and let `p` follow regular typography guides.

src/pages/faq.astro
Anca2022: The `<summary>` element has a padding-bottom: 2xl when closed indeed, but that changes to padding-bottom: md when expanded. ~This difference in padding causes a slight visual flicker on the `<summary>` during the open/close transition~.
To fix this, I separated the padding between the two elements

src/pages/faq.astro
Anca2022: There’s only one example where an answer contains multiple paragraphs (for the question “What is a wallet address”). In that case, I needed spacing between the paragraphs themselves, as well as spacing between the end of the answer and the `<details> `container.
I could move the padding to the wrapper `<div>`, but I’d still need to handle the spacing between the `<p>` elements inside the answer.

sidvishnoi: I'd recommend that. So the `<div>` is concerned with overall layout padding, and we can edit spacing between paragraphs like a regular copy thing.

src/pages/faq.astro
Anca2022: My bad, that was not it
changing the padding back to `<summary>`

src/pages/faq.astro
```diff
@@ -175,18 +171,24 @@ const t = useTranslations(lang)
       font: var(--font-h2);
     }
 
-    details.faq-item {
-      padding: var(--padding-2xl);
+    details.faq-item summary {
+      padding: var(--padding-2xl) calc(2 * var(--padding-2xl)) var(--padding-2xl)
+        var(--padding-2xl);
```
sidvishnoi:
```suggestion
      padding: var(--padding-2xl);
      padding-right: calc(2 * var(--padding-2xl));
```

sidvishnoi (approved):

PR #608, opened 2025-11-10
fix(examples/counter): use the `Intl.NumberFormat` API

tomayac:
Rather than hardcode the currency, use the `Intl.NumberFormat` API for this.

### DIFF
--- a/public/interactive-examples/wm2-count-revenue/counter.js
+++ b/public/interactive-examples/wm2-count-revenue/counter.js
@@ -12,7 +12,7 @@ if (window.MonetizationEvent) {
 
     total += Number(value);
 
-    contribution.innerText = new Intl.NumberFormat('en-US', {
+    contribution.innerText = new Intl.NumberFormat(navigator.language, {
       style: 'currency',
       currency,
     }).format(total);


public/interactive-examples/wm2-count-revenue/counter.js
```diff
@@ -2,19 +2,19 @@ let total = 0;
...
     console.log(`Browser sent ${currency} ${value}.`);
 
-    if (total === 0) {
-      document.getElementById("currency").innerText = currency;
-    }
-
     total += Number(value);
 
-    document.getElementById("total").innerText = total.toFixed(9);
+    contribution.innerText = new Intl.NumberFormat('en-US', {
```
sidvishnoi: So that it can use whatever default locale the user has, not just en-US:
```suggestion
    contribution.innerText = new Intl.NumberFormat([], {
```

public/interactive-examples/wm2-count-revenue/index.html
```diff
@@ -3,13 +3,12 @@
...
-  <script type="application/javascript" src="./counter.js"></script>
+  <script defer type="application/javascript" src="./counter.js"></script>
 </head>
 
 <body>
   <p>
     Thanks for contributing
-    <span id="total">0.00 USD</span>
-    <span id="currency"></span>
+    <span id="contribution">maybe in the future</span>.
```
sidvishnoi, line 12: @melissahenderson suggestions for a better default text? It says: "Thanks for contributing maybe in the future" here.

public/interactive-examples/wm2-count-revenue/counter.js
```diff
@@ -2,19 +2,19 @@ let total = 0;
...
+      style: 'currency',
+      currency,
+    }).format(total);
```
sidvishnoi: While it's true the spec uses 3-letter ISO country codes, some niche Web Monetization wallets are using custom 5-letter codes (for non-standard virtual currencies). So, using `Intl.NumberFormat` will crash there (it gracefully handles any 3-letter code, even if not a standard one).
Not sure if it's a good idea to let this complication be handled here (to prevent uncaught exception in event callback), but leaving as a note for future.

sidvishnoi (commented):
Thanks @tomayac!

public/interactive-examples/wm2-count-revenue/index.html
melissahenderson: I think we should take out the hardcoded text and make the entire message change depending on whether the visitor is web monetized. Is that possible?

- Thanks for contributing $0.00.
- Consider supporting me with Web Monetization.

Or something like that.

public/interactive-examples/wm2-count-revenue/counter.js
tomayac:
```suggestion
    contribution.innerText = new Intl.NumberFormat(navigator.language, {
```

tomayac: I'm using `navigator.language` now, which should always work.

public/interactive-examples/wm2-count-revenue/counter.js
tomayac: These wallets are in [violation of the spec](https://webmonetization.org/specification/#currency-member) then, which _requires_ the [`currency` member to be a well-formed 3-letter alphabetic code](https://www.w3.org/TR/payment-request/#dom-paymentcurrencyamount-currency). I'd acknowledge the fact in the docs as a pitfall, but not complicate the illustrative code sample.

public/interactive-examples/wm2-count-revenue/index.html
tomayac: I don't think this is necessary as it complicates the code sample. IMHO _"Thanks for contributing maybe in the future"_ seems like a basic compromise that people actually implementing this can extrapolate from and put whatever they see fit, like a CtA to install the extension, donate through other means, or just not show anything. Alternatively, _"Thanks for contributing $0.00."_ seems passive-aggressive to me, plus, ideally, would also need adjusting the currency, but it's not known. I won't die on this hill, though, but, I think, the example is good to go as is. Your call :-)

sidvishnoi: +1 to not complicate the code sample.
We can look into a better default message later.

PR #617, opened 2025-11-28
Remove Umami event attributes from site links, add the ones bibi wanted

JonathanMatthey:
#### Description

- Removed all data-umami-event tags from navigation, footer, hero CTAs, carousel controls/cards, and tool/link lists.
- Cleaned localized HTML strings and FAQ answers so links no longer carry Umami tracking attributes.
- Ensured wallets, updates, and other homepage/pillar page links no longer include tracking tags in their markup; source now shows plain anchor tags.
- add specific umami events, navigation and homepage

#### Testing

- pnpm start
- Manually view page source in the running site to confirm links no longer include data-umami-event attributes.


src/components/pages/PluginList.astro
```diff
@@ -35,7 +35,6 @@ const cardList: Plugin[] = [...cards.slice(0, 3)]
           <a
             href={card.cta.href}
             class='btn flexible secondary'
-            data-umami-event={card.cta.event}
```
Anca2022, line 41: we should also remove the `event` property from `plugins` objects in `/src/data/developers.ts`, since it’s no longer being used.

Anca2022: And, in general, we should remove this property everywhere else it’s no longer being used.

Anca2022 (commented):
Just the `event` properties, the rest looks good! 😃

src/components/pages/WalletList.astro
JoblersTune, line 1: I would have thought we would be interest in wallet list clicks, especially for the Interledger app wallet.

src/components/pages/AudienceHero.astro
```diff
@@ -45,14 +45,13 @@ import PillTag from './PillTag.astro'
       <a
         href={content.primaryCta.href}
         class='btn primary'
-        data-umami-event={content.primaryCta.event}>{content.primaryCta.text}</a
```
JoblersTune, line 48: I'm surprised that we'd remove event tracking for the primary audience CTAs.

src/pages/index.astro
```diff
@@ -126,7 +123,6 @@ const t = useTranslations(lang)
           <a
             class='btn primary flexible'
             href='/publishers'
-            data-umami-event='Home page link - Publishers'
```
JoblersTune: These are also primary CTAs, why would we not care about them being clicked?

JoblersTune:
Reviewers need to check this is in line with what was laid out in [WEB-94](https://linear.app/interledger/issue/WEB-94/clean-up-event-tracking-for-webmon-keep-it-simple )

melissahenderson:
@JonathanMatthey @JoblersTune This is from November 2025. Can we close it?

PR #622, opened 2025-12-09
chore: updated favicon with an ico variant

JoblersTune:
image

Moving to ico as a fully supported favicon image format. However, reading Google's guidelines I can't figure out why the svg was not dispalying though. See [https://developers.google.com/search/docs/appearance/favicon-in-search](https://developers.google.com/search/docs/appearance/favicon-in-search)


public/favicon.svg
sidvishnoi, line 1: I would recommend keeping both.

JoblersTune: Ok, when I read "Google Search only supports one favicon per site" I worried having two listed might not be best. I'll add the other back in. DO you suggest the svg first and the ico file as second in line?

sidvishnoi: As .ico is fallback, it would make sense to add it last, but I've seen both the orders in practice.
Also, SVG favicons can be shown in browsers (not just search engines) more clearly than .ico ones (think recently visited tabs, shortcuts etc.).

PR #627, opened 2026-01-14
docs: add csp info to pub tools and other updates

melissahenderson:
This update started out as an update to the banner and widget pages, by adding CSP info. (https://github.com/interledger/publisher-tools/pull/515)

While making these updates and clicking through my local version, there were a handful of pages that would break. Long story short, the Steps component wasn't playing nicely with asides. I removed all instances where the Step component was used. Doing this also resolved a few other issues w/ tabs and the LinkOut component.

I also made content updates based on Vale feedback, removed `{/* prettier-ignore */}` from places where it's no longer needed, and removed the usage of the Disclosure tag. We stopped using it a while ago, but forgot to change the page.

### DIFF
--- a/src/content/docs/publishers/banner-tool.mdx
+++ b/src/content/docs/publishers/banner-tool.mdx
@@ -63,6 +63,10 @@ All pages that include the banner script are automatically web monetized with yo
 
 All pages that include the script are now web monetized with the new wallet address.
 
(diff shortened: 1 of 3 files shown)


melissahenderson:
Forgot about partials and Prettier. Fixing...

PR #634, opened 2026-01-21
Update Chimoney App availability to Canada and Mexico

edent:
Just spoke with Uchi Uchibeke. Chimoney only officially supports Canada and Mexico. Interledger wallets not available in other countries.

PR #642, opened 2026-02-16
fix(/wallets): update Chimoney App availability to Canada and Mexico

sidvishnoi:
Reopens https://github.com/WICG/webmonetization/pull/634 from @edent.

---

Just spoke with Uchi Uchibeke. Chimoney only officially supports Canada and Mexico. Interledger wallets not available in other countries.


uchibeke (approved):
LGTM ♠️

PR #644, opened 2026-02-19
fix(spec/flows): fix ReSpec errors

sidvishnoi:
Fix remaining (13) ReSpec linking errors in flows spec.

### DIFF
--- a/specification/flows/section/wallet-setup.html
+++ b/specification/flows/section/wallet-setup.html
@@ -17,7 +17,7 @@
     </dd>
     <dt><dfn>Interactive grant</dfn></dt>
     <dd>
-      Open Payments requires outgoing payment [=grant=] requests to be interactive. When a grant request is interactive, it means explicit interaction by an individual (typically the client’s end user) is a required step in the delegation process.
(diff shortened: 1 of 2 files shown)


specification/flows/section/wallet-setup.html
```diff
@@ -17,7 +17,7 @@ <h3>Terms</h3>
     </dd>
     <dt><dfn>Interactive grant</dfn></dt>
     <dd>
-      Open Payments requires outgoing payment [=grant=] requests to be interactive. When a grant request is interactive, it means explicit interaction by an individual (typically the client’s end user) is a required step in the delegation process.
+      Open Payments requires outgoing payment [=grants|grant=] requests to be interactive. When a grant request is interactive, it means explicit interaction by an individual (typically the client’s end user) is a required step in the delegation process.
```
asurkov: not sure if plural is the right change here

asurkov (approved):

specification/flows/section/wallet-setup.html
sidvishnoi: `outgoing payment grant requests` (as above) vs `outgoing payment grants requests`. Maybe we can change the `Grant` definition. Then we can link it either as grants or grant.

PR #649, opened 2026-03-19
chore(docs): move partials files for Astro/Starlight

melissahenderson:
This issue partly resolves the tech writers' DOC-24 Linear issue

Previously, the partials subfolder was outside of the docs subfolder. Astro/Starlight made a change where asides used in partials would no longer render if the partials files were outside of the docs folder. 

This issue moves the partials so that we don't have issues in the future.

### DIFF
--- a/src/content/docs/publishers/widget-tool.mdx
+++ b/src/content/docs/publishers/widget-tool.mdx
@@ -4,7 +4,7 @@ title: Widget tool
 
 import { LinkButton } from '@astrojs/starlight/components'
 import { LinkOut } from '@interledger/docs-design-system'
-import CSP from './partials/_csp.mdx'
+import CSP from '/src/content/docs/partials/_csp.mdx'
 
 Add a <LinkOut href="https://webmonetization.org/tools/widget">payment widget</LinkOut> to your site so visitors can send you a one-time payment. The widget appears to all visitors, even those with an active Web Monetization extension.
 


src/content/docs/developers/link-element-webpage.mdx
```diff
@@ -3,19 +3,19 @@ title: Webpage (HTML)
...
-import MultiLink from '/src/partials/link-element-multilink.mdx'
-import MultiLinkTip from '/src/partials/link-element-multilink-tip-text.mdx'
-import Iframes from '/src/partials/link-element-iframes.mdx'
-import IframesNote from '/src/partials/link-element-iframes-note-text.mdx'
-import Media from '/src/partials/link-element-media.mdx'
-import HowHtml from '/src/partials/link-element-how-html.mdx'
-import HowRel from '/src/partials/link-element-how-rel.mdx'
-import HowAgent from '/src/partials/link-element-how-agent.mdx'
-import ContPayments from '/src/partials/link-element-cont-payments.mdx'
+import Overview from '/src/content/docs/partials/link-element-overview.mdx'
```
sidvishnoi: Would it be helpful to add an alias to partials dir? Then we can have imports like:
```js
import Overview from '#partials/link-element-overview.mdx'
```

melissahenderson: Could we set something like that up later? I need to do the same thing with Rafiki, Open Payments, and maybe another I'm forgetting at the moment.

PR #653, opened 2026-03-26
feat(spec/ext): add Extensions API specification

asurkov:


package-lock.json
sidvishnoi, line 1: Can you please reset changes to this file? We try to not touch lock files unless package.json is modified.

specification/extensions-api/index.html
```diff
@@ -0,0 +1,92 @@
...
+        "infra",
+        "webidl"
+      ],
+    };
+  </script>
+</head>
+<body>
+<section id="abstract">
+  <p>
+    This specification defines an extensions API for managing a browser's
```
sidvishnoi:
```suggestion
    This specification defines a browser extension API for managing a browser's
```

specification/extensions-api/section/events.html
```diff
@@ -0,0 +1,80 @@
...
+    Events in this namespace are scoped to a
+    <a data-cite="HTML/document-sequences.html#navigable">document</a>,
+    identified by its tab and frame. Each event callback receives a
+    {{MonetizationInfo}} dictionary describing the affected document.
+  </p>
+
+  <section data-dfn-for="monetization">
+    <h3><dfn>onStarted</dfn></h3>
+
+    <pre>monetization.onStarted.addListener(
```
sidvishnoi:
```suggestion
    <pre class="example">browser.monetization.onStarted.addListener(
```

specification/extensions-api/section/events.html
```diff
@@ -0,0 +1,80 @@
...
+            monetization started.</dd>
+        </dl>
+      </dd>
+    </dl>
+  </section>
+
+  <section data-dfn-for="monetization">
+    <h3><dfn>onStopped</dfn></h3>
+
+    <pre>monetization.onStopped.addListener(
```
sidvishnoi:
```suggestion
    <pre>browser.monetization.onStopped.addListener(
```

specification/extensions-api/section/interfaces.html
```diff
@@ -0,0 +1,18 @@
+<section id="interfaces">
+  <h2>The <code>monetization</code> namespace</h2>
```
sidvishnoi, line 2: Is this available only to background/service worker, or popup/content scripts as well?

specification/extensions-api/section/types.html
```diff
@@ -0,0 +1,38 @@
...
+      <dt><dfn>frameId</dfn></dt>
+      <dd>
+        The identifier of the frame within the tab.
+        <code>0</code> denotes the top-level (main) frame.
+      </dd>
+
+      <dt><dfn>url</dfn></dt>
+      <dd>
+        The URL of the monetized document.
+      </dd>
```
sidvishnoi, line 35: 1. How is this intended to be used? 
2. Is this frame URL or tab URL?
3. This inherently exposes some details that require `tabs` permission. If this is meant to be `tab.url`, should we defer that to existing `tabs` permission?

specification/extensions-api/section/events.html
```diff
@@ -0,0 +1,80 @@
...
+    <h3>Event ordering</h3>
+
+    <p>
+      For a given document, an {{monetization/onStarted}} event is always
+      delivered before any corresponding {{monetization/onStopped}} event.
+      If a document has multiple concurrent sessions, intermediate
+      session starts and stops do not produce additional events.
+      Only the transition from zero to one active session
+      ({{monetization/onStarted}}) and from one to zero active sessions
+      ({{monetization/onStopped}}) fire events.
```
sidvishnoi, line 76: 👍

specification/extensions-api/section/events.html
```diff
@@ -0,0 +1,80 @@
+<section id="events">
+  <h2>Events</h2>
+
+  <p>
+    Events in this namespace are scoped to a
+    <a data-cite="HTML/document-sequences.html#navigable">document</a>,
```
sidvishnoi:
```suggestion
    [=navigable=],
```

sidvishnoi: Or do you mean `{{ Document }}` only?

specification/extensions-api/section/interfaces.html
asurkov: it's scoped to privileged content so should be available in background/service workers and in popup/options pages, but not available in content scripts. What scope do you think should be?

specification/extensions-api/section/types.html
asurkov: That's a bigger question which we probably should move out of context of this starter doc, now it just my match the implementation we have. I started a new issue https://github.com/WICG/webmonetization/issues/655

specification/extensions-api/section/events.html
asurkov: {{ Document }} you're right.

specification/extensions-api/section/interfaces.html
sidvishnoi: Everything except content script seems good to me.
Both background and popup can have use cases for this API, so limiting just to background won't be beneficial either.

asurkov: I didn't catch: do you think we need it available in content scripts too or it's not really necessary?

sidvishnoi: I mean, I agree with what you said in https://github.com/WICG/webmonetization/pull/653#discussion_r3015462530: keeping for background and popup/pages. Excluding content scripts.

package-lock.json
sidvishnoi: I think we can merge this PR with this change.

sidvishnoi (approved):
:shipit:

PR #664, opened 2026-05-01
feat(spec/ext) specify getWallet/onPayment

asurkov:


specification/extensions-api/section/types.html
```diff
@@ -1,18 +1,91 @@
...
+      </dd>
+
+      <dt><dfn>name</dfn></dt>
+      <dd>
+        A human-readable name for the wallet.
+      </dd>
+
+      <dt><dfn>currency</dfn></dt>
+      <dd>
+        An ISO 4217 currency code (e.g. <code>"USD"</code>).
```
meyerweb: Not that this needs to be changed immediately, but there is the possibility of loosening the currency code requirements to be a string, not necessarily conforming to any specific ISO or other standard, though with ISO 4217 as a default.  See https://github.com/WICG/webmonetization/pull/651 for a proposed change (which is apparently still being debated at Interledger).

specification/extensions-api/section/types.html
```diff
@@ -1,18 +1,91 @@
...
+      </dd>
+
+      <dt><dfn>initial</dfn></dt>
+      <dd>
+        A decimal string representing the total budget for the current
```
meyerweb: Should this be in scaled currency (e.g., `1000`) instead of a decimal?

specification/extensions-api/section/types.html
```diff
@@ -1,18 +1,91 @@
...
+        billing period (e.g. <code>"10.00"</code>).
+      </dd>
+
+      <dt><dfn>remaining</dfn></dt>
+      <dd>
+        A decimal string representing the remaining budget for the
```
meyerweb: Ditto to line 52.

meyerweb (commented):
Some thoughts on first read.

specification/extensions-api/section/events.html
```diff
@@ -67,13 +66,40 @@ <h4>Parameters</h4>
...
+
+    <p>
+      Fired when a payment is sent for a monetization session. The
+      updated budget balance can be retrieved via
+      {{monetization/getWallet()}}.
+    </p>
+
+    <h4>Parameters</h4>
+    <dl>
+      <dt><code>callback</code></dt>
```
sidvishnoi, line 93: Nit: ReSpec supports this syntax for `<code>`
```suggestion
      <dt>`callback`</dt>
```

specification/extensions-api/section/types.html
```diff
@@ -1,18 +1,91 @@
...
-      Describes the document where a monetization state change occurred.
+      Describes the connected user wallet, including its current
+      budget balance. Currency amounts follow
+      {{CurrencyAmount}} conventions defined in [[WebMonetization]].
     </p>
 
     <pre class="idl">
-      dictionary MonetizationInfo {
+      dictionary Wallet {
+        required DOMString address;
```
sidvishnoi, line 26: Is this the `address` user provided (e.g. `https://sidvishnoi.com/.well-known/pay`) or the resolved wallet address ID (e.g. `https://ilp.gatehub.net/981946513/eur`)?

specification/extensions-api/section/types.html
```diff
@@ -1,18 +1,91 @@
...
+        required DOMString name;
```
sidvishnoi, line 27: In Open Payments, the `publicName` field is optional.

specification/extensions-api/section/types.html
```diff
@@ -1,18 +1,91 @@
...
+        required DOMString currency;
+        required DOMString initial;
+        required DOMString remaining;
```
sidvishnoi: Should these have "balance" in their names?

specification/extensions-api/section/types.html
```diff
@@ -1,18 +1,91 @@
...
+        DOMString? renewDate;
+        required boolean needsReconnect;
```
sidvishnoi: I think we discussed we might want a "reason" along with `needsReconnect`?

specification/extensions-api/section/types.html
sidvishnoi: We're aligning this more with `PaymentCurrencyAmount` and `Intl` APIs, than with Open Payments.

sidvishnoi: Let's mention these are compatible with `PaymentCurrencyAmount` or link to PaymentRequest/WebMonetization spec

specification/extensions-api/section/types.html
```diff
@@ -1,18 +1,91 @@
...
+      <dt><dfn>renewDate</dfn></dt>
+      <dd>
+        An ISO 8601 date string indicating when the budget next renews.
+        Absent when the budget does not renew monthly.
+      </dd>
+
+      <dt><dfn>needsReconnect</dfn></dt>
+      <dd>
+        <code>true</code> when the stored access token is invalid and the
+        wallet must be reconnected.
```
sidvishnoi: What happens when remaining budget is too low? Do we set `needsReconnect` then as well?

specification/extensions-api/section/types.html
sidvishnoi: Maybe `name` will be empty string in those cases?

specification/extensions-api/section/types.html
asurkov: I changed it referring to PaymentCurrencyAmount same as initial/remaining fields.

specification/extensions-api/section/types.html
asurkov: address id, I tweaked words to make that explicit

specification/extensions-api/section/types.html
asurkov: I think it's a good idea

specification/extensions-api/section/types.html
asurkov: good point, changed that to optional reconnectReason

specification/extensions-api/section/types.html
asurkov: nothing right now I suspect, needsReconnect is set on the user wallet response, current implementation doesn't guess about reasons.

sidvishnoi (approved):

specification/extensions-api/section/types.html
asurkov: right, it can be an empty string if not provided

PR #678, opened 2026-07-20
chore(spec): format spec and add format command

MichaelKohler:
This removes the respec script and GitHub action workflow, and introduce a `format:spec` script and documentation for it. I then also ran the formatting script which resulted in minor changes to the spec format. It initially also wanted to remove the comments we have, so I disabled that in the config.

Note that this currently only formats the WM spec, but not the WM Flows and WM Extension API spec. Those will need a bit more work to properly handle the includes. And then we can also add it as a check step in the GitHub Action workflows.


specification/index.html
```diff
@@ -739,10 +738,10 @@ <h2>
       <aside class="note">
         <p>
           The <a href=
-          "https://html.spec.whatwg.org/#attr-iframe-allow">allow</a>
-          attribute only takes effect when the navigable content of the
-          [^iframe^] is navigated. Adding or removing the monetization
-          attribute has no effect on an already-loaded document.
+          "https://html.spec.whatwg.org/#attr-iframe-allow">allow</a> attribute
```
sidvishnoi: While we're here...

```suggestion
          The [^iframe/allow^] attribute
```

sidvishnoi (approved):

PR #695, opened 2026-08-11
feat(spec): add examples for Permission Policy

MichaelKohler:
Adding two examples for the Permission Policy, covering both `allow` attribute and `Permission-Policy` header usage.

### DIFF
--- a/specification/index.html
+++ b/specification/index.html
@@ -742,6 +742,27 @@
           monetization attribute has no effect on an already-loaded document.
         </p>
       </aside>
+      <aside class="example">
+        <p>
+          Given a page embedding the following [^iframe^], monetization is
+          allowed for the embedded <a>document</a>:
+        </p>
+        <pre class="html">
+          &lt;iframe src="https://example.com/" allow="monetization"&gt;&lt;/iframe&gt;
+        </pre>
+      </aside>
+      <aside class="example">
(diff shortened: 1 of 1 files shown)


specification/index.html
meyerweb, line 797:
```suggestion
            URL provided does not match monetization-src's source list:
```

meyerweb, line 796:
```suggestion
            Fetches for the following code will return a network error, as the
```

meyerweb (changes requested):
Noticed a couple of grammatical errors nearby while reviewing the changes.

specification/index.html
meyerweb: That last edit could also be “Fetches for the following code will return network errors, as the” depending on how you want to shade it (and especially if a single fetch can return multiple errors).

MichaelKohler: Good catch! I've created a separate PR for that as it's unrelated and I like to keep PRs focused on one topic: https://github.com/WICG/webmonetization/pull/715

MichaelKohler: Good catch! I've created a separate PR for that as it's unrelated and I like to keep PRs focused on one topic: https://github.com/WICG/webmonetization/pull/715

sidvishnoi (approved):

PR #697, opened 2026-08-12
feat(spec): add `onmonetization` IDL spec

### LINKED ISSUE #696: Write spec for onmonetization
Currently `onmonetization` is defined in prose, but not in IDL. We should write the IDL spec outside the monkey patched section.

MichaelKohler:
Closes #696.

This PR introduces the `onmonetization` IDL specification as currently it is only defined in prose. It also aligns the specification's event model with how Web Monetization is actually implemented and used in practice, as before it was not defined for any `Document` or `Window`.

And finally, I have reworded the event firing specification to better define the event to be used, as before it could be understood that `event` is not used.

### DIFF
--- a/specification/index.html
+++ b/specification/index.html
@@ -503,46 +503,65 @@
     </section>
     <section>
       <h2>
-        `onmonetization` event handler
+        Events
       </h2>
-      <p>
-        The `onmonetization` event handler is an [=event handler content
-        attribute=] that can be applied to any [=element=]. The user agent uses
-        it to notify that some [^link^] has been [=monetized=].
-      </p>
-    </section>
-    <section id="events">
-      <h2>
-        monetization event
-      </h2>
-      <p>
-        When the [=payment session=] has sent a |payment| with a non-zero
-        amount, perform the following steps:
-      </p>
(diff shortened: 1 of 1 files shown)


specification/index.html
```diff
@@ -503,46 +503,65 @@ <h2>
...
-        </li>
-      </ol>
+      <section data-dfn-for="GlobalEventHandlers">
+        <h3>
+          Extensions to `GlobalEventHandlers` interface
+        </h3>
+        <p>
+          The <code><dfn data-cite=
+          "html/webappapis.html#globaleventhandlers">GlobalEventHandlers</dfn></code>
+          interface is defined in [[HTML]].
```
sidvishnoi: Does this give any ReSpec errors?
```suggestion
          The {{GlobalEventHandlers}} interface is defined in [[HTML]].
```

MichaelKohler: Nope, it does not. Good point. Also took this chance to also update the format of the spec.

sidvishnoi (approved):

asurkov (approved):
lgtm

PR #700, opened 2026-08-17
chore(ci): granular GitHub workflow permissions and use env variables

MichaelKohler:
Adds an explicit least-privilege `permissions:` block to the workflows that previously had none. Each is scoped to only what its steps actually use.

Routes `${{ }}` values through `env` instead of interpolating them directly.


.github/workflows/contributors.yml
sidvishnoi, line 1: Can remove this workflow. It's not used anymore. There's no `contributors.json` either.

sidvishnoi (approved):

PR #705, opened 2026-08-20
fix(docs): Adjust card colors in dark mode

MichaelKohler:
Currently the announcement card on https://webmonetization.org/docs/ and https://webmonetization.org/resources/get-involved/ is too bright in dark mode and makes the links hard too see.

This adjusts the colors to use the same colors as the "note" on https://webmonetization.org/supporters/get-started/.


sidvishnoi (approved):
Good catch!

MichaelKohler:
Kudos goes to Jessica who found it.

PR #706, opened 2026-08-21
fix(playground): allow duplicate wallet addresses and fix "Enabled" toggle

### LINKED ISSUE #493: Playground: allow adding duplicate link tags
As mentioned in https://github.com/WICG/webmonetization/pull/492#issuecomment-2591829958 - browser allows duplicate link tags, so playground shouldn't add that restriction artificially.

MichaelKohler:
Closes #493.

This also fixes the "Enabled" toggle.

### DIFF
--- a/src/pages/play.astro
+++ b/src/pages/play.astro
@@ -265,12 +265,6 @@ const t = useTranslations(lang)
 
     const walletAddressUrl = toWalletAddressUrl(walletAddress)
 
-    if (addressAlreadyAdded(walletAddressUrl)) {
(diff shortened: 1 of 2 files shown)


sidvishnoi (changes requested):
This is not enough 😄 
Events are attached to first address in the playground - they should be separate for each link tag.

MichaelKohler:
> Events are attached to first address in the playground - they should be separate for each link tag.

Yep, my testing was bad for this one, sorry!

sidvishnoi (approved):

PR #712, opened 2026-09-03
chore(deps): upgrade dependencies for Astro 7 (DOCSDEV-75)

Infi-Knight:
This one moves the site to Astro 7.

## Dependency changes

- `astro` 6.3.8 to 7.2.10
- `@astrojs/starlight` 0.39.2 to 0.42.0
- `@astrojs/markdown-remark` added at 7.3.0
- `@interledger/docs-design-system` 0.13.0 to 0.14.0
- `starlight-links-validator` 0.24.0 to 0.26.0
- `starlight-llms-txt` 0.9.0 to 0.11.0
- `@astrojs/check` 0.9.9 to 0.9.10
- `sharp` 0.34.5 to 0.35.4
- `respec` 35.6.1 to 37.3.5

Tooling: `eslint` to 10.9.1, `eslint-plugin-astro` to 3.1.0, `astro-eslint-parser` to 3.1.0, `typescript-eslint` to 8.69.0, `globals` to 17.11.0, `prettier` to 3.9.6, `eslint-plugin-jsx-a11y` added at 6.10.2.

`typescript` stays at 6.0.3. `@astrojs/check` declares `typescript ^5 || ^6`, and `typescript-eslint` refuses TypeScript 7 at runtime.

## Removed `remark-mermaidjs`

It was declared in `dependencies` but never imported. A repo-wide search found references only in `package.json`, the lockfile, and one old SVG filename. This site renders no Mermaid diagrams.

## An accessibility bug, found and fixed

`eslint-plugin-astro` 3 exposes accessibility rule sets that version 1 did not. The config now extends `flat/jsx-a11y-recommended`, which turns on 31 rules. That found one real problem:

```
src/pages/faq.astro
  14:7  error  img elements must have an alt prop  astro/jsx-a11y/alt-text
```

The FAQ hero image had no `alt`. It sits beside the "Frequently asked Questions" heading and adds no information, so it is decorative and now has `alt=""`. Screen readers will skip it rather than announce a filename.

## Two things reviewers should know

**`package-lock.json` is fully regenerated.** npm could not resolve the Astro 7 tree against the old lockfile, so the diff is large. I verified the Linux-only optional dependencies are still recorded, so `npm ci --include=optional` still works on CI. Note that Vite 8 replaced rollup with `rolldown`, so the pinned `@rollup/rollup-linux-x64-gnu` is now unused. I left it in place rather than change a deliberate pin, but it can be dropped in a follow-up.

**A new `overrides` entry was needed.** `eslint-plugin-jsx-a11y@6.10.2` declares `eslint ^3 || ... || ^9` but runs correctly on ESLint 10. npm treats that stale range as a hard install failure, where bun and pnpm only warn. The override forces the root `eslint` version for that package. It can be removed once upstream declares ESLint 10 support.

## Markdown processor

Astro 7 makes Sätteri the default, and Sätteri does not run remark or rehype plugins. Starlight follows the same default. This cycle keeps every Starlight site on the `unified()` processor, so all sites stay on the pipeline they run today. This site configures no plugins, so its `unified()` call takes no arguments.

## Test plan

- [x] `npm install --include=optional` clean
- [x] `npm run build` passed, 41 pages, 101 HTML files
- [x] **The `spec-symlink` integration works under Vite 8.** Tested from a clean state with `public/specification` deleted, which is what CI does. The integration recreated the symlink and 22 spec files reached `dist/`.
- [x] `/specification/` serves, and respec loads from the W3C CDN (`respecConfig` defined, 23825 characters of content)
- [x] `starlight-links-validator` 0.26.0 reports all internal links valid
- [x] `/llms.txt` and `/llms-full.txt` generate
- [x] `npm run lint` exits 0 at `--max-warnings=0` after the alt-text fix
- [x] Pagefind index built
- [x] Linux-only optional dependencies still present in the regenerated lockfile
- [ ] Deploy preview verified, including `/specification/`

## Two pre-existing build warnings, both benign

- `[starlight-i18n-loader] No files found matching ... in "src/content/i18n"`. That directory holds only a `.keepme` file. No i18n is configured.
- `Could not render /404 from route /[...slug] as it conflicts with higher priority route /404`. `dist/404.html` is still produced.

Refs: DOCSDEV-75

### DIFF
--- a/astro.config.mjs
+++ b/astro.config.mjs
@@ -3,6 +3,7 @@ import path from 'node:path'
 import { symlink } from 'node:fs/promises'
 import { fileURLToPath } from 'node:url'
 import starlight from '@astrojs/starlight'
+import { unified } from '@astrojs/markdown-remark'
 import starlightLinksValidator from 'starlight-links-validator'
 import starlightFullViewMode from 'starlight-fullview-mode'
 import starlightLlmsTxt from 'starlight-llms-txt'
@@ -10,6 +11,9 @@ import starlightLlmsTxt from 'starlight-llms-txt'
 // https://astro.build/config
 export default defineConfig({
   site: 'https://webmonetization.org',
+  markdown: {
+    processor: unified()
+  },
   integrations: [
     starlight({
       title: 'Web Monetization',
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -13,6 +13,7 @@ export default defineConfig([
   js.configs.recommended,
   tseslint.configs.recommended,
   eslintPluginAstro.configs.recommended,
+  eslintPluginAstro.configs['flat/jsx-a11y-recommended'],
   globalIgnores(['dist', '.astro', 'node_modules', 'public', '**/*.min.js']),
   {
     files: ['**/*.{js,mjs,cjs}'],
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
         "@astrojs/check": "^0.9.10",
         "@astrojs/markdown-remark": "^7.3.0",
         "@astrojs/starlight": "^0.42.0",
-        "@interledger/docs-design-system": "^0.14.0",
+        "@interledger/docs-design-system": "^0.14.2",
         "astro": "~7.2.10",
         "respec": "^37.3.5",
         "sharp": "^0.35.4",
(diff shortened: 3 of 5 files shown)


Anca2022 (approved):
lgtm 😃 

Same case here - needs to wait for `interledger/docs-design-system` to be bumped before it’s complete.

PR #724, opened 2026-09-21
feat(spec): Fix links and clarify behavior for link tag

### LINKED ISSUE #698: Finalize spec for link lifecycle
As per this (old) [comment](https://chromium-review.googlesource.com/c/chromium/src/+/4930834/11..83/third_party/blink/renderer/core/html/link_monetization.cc#b45) from dbaron, we should define steps for the `link` tag that tie properly into https://html.spec.whatwg.org/multipage/semantics.html#fetching-and-processing-a-resource-from-a-link-element.

We need to check what is already done and what still needs to be done.

At the same time, we should fill the gaps in the lifecycle of the tag, as described in https://docs.google.com/document/d/1-M3cjFGH8EluL9KNMzDcdRzNzSLnc83cRaqudPBRhaI/edit?tab=t.0#bookmark=id.ocoxu518ieg9:

> It enumerates its own fetch-and-process triggers (creation, becoming browsing-context connected, and changes to href, disabled, crossorigin or type) and states one MUST: “If a ‘monetization’ link is no longer browsing-context connected, a user agent MUST stop the payment session.”
It does not state what happens to a session already running when one of those triggers fires, and rel is not among the triggers.

Note: for load/error events there are quite a few discussions open in this repo, see https://github.com/WICG/webmonetization/issues?q=state%3Aopen%20label%3A%22link%22. Likely out of scope of this issue, but keep it in mind that this is yet another topic to be tackled.

MichaelKohler:
The WM spec steps already tie in nicely with the HTML spec, so there is nothing to be done in that regard. However there are a few more improvements needed to finalize the link tag spec:

* Added explicit step to stop payment sessions
* Added explicit handling of disabled attribute
* Added sentence to declare monetization keyword body-ok.
* Clarified payment session creation condition to not create a payment session on failure (found by Copilot in this PR's review).
* Fixed links to work in our specification (it was previously copied from the WHATWG spec). We will need to adjust it again once moving over to the HTML spec, but for now I prefer working links.
* Cleaned up initial paragraphs of link tag to not repeat itself so much.
* Removed unnecessary div
* Formatted spec (change in CSP section)

Note on `rel`: this does not seem to be explicitly specified by any other keywords either, and should already be covered by “When the external resource link is created on a link element that is already browsing-context connected.”

Closes #698.


specification/index.html
```diff
@@ -300,204 +300,190 @@ <h2>
         </p>
       </aside>
       <p>
-        The monetization keyword indicates a [=wallet address=] used to
-        monetize the document.
+        The <code>[=monetization=]</code> keyword indicates a [=wallet
```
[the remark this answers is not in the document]
MichaelKohler: It does.

specification/index.html
```diff
@@ -896,24 +882,28 @@ <h4>
...
-                </ins>
+                <dt>
+                  ... <ins>
+                  <dt>
+                    `monetization-src`
+                  </dt>
+                  <dd>
+                    `&lt;&lt; "monetization-src", "default-src" &gt;&gt;`
+                  </dd></ins>
+                </dt>
```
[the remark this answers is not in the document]
MichaelKohler: Will create a separate PR for this.

specification/index.html
```diff
@@ -300,204 +300,190 @@ <h2>
...
-        monetize the document.
+        The <code>[=monetization=]</code> keyword indicates a [=wallet
+        address=] used to monetize the document. It indicates that some aspect
+        of the document is [=monetized=].
       </p>
       <p>
-        The <code data-x="rel-monetization">monetization</code> keyword may be
-        used with <code>link</code> elements. This keyword creates an
-        <span data-x="external resource link">external resource link</span>.
+        The <code>monetization</code> keyword may be used with [^link^]
```
sidvishnoi: Nit: can use backticks for `<code>`
~~~suggestion
        The `monetization` keyword may be used with [^link^]
~~~

MichaelKohler: Applied it to all.

specification/index.html
```diff
@@ -896,21 +879,24 @@ <h4>
             Get fetch directive fallback list
           </h4>
           <p>
-            This spec patches the [[CSP]] specification's <a data-cite="CSP#directive-fallback-list">directive fallback list</a>
+            This spec patches the [[CSP]] specification's <a data-cite=
```
MichaelKohler, line 878: I have moved this formatting to https://github.com/WICG/webmonetization/pull/725 and will rebase this PR here once merged.

specification/index.html
```diff
@@ -300,204 +300,187 @@ <h2>
...
+          <p>
+            If <var>element</var>'s [^link/disabled^] attribute is set, stop
+            all previous [=payment sessions=] initiated by |element|'s
+            [=external resource link=], then return false.
+          </p>
+        </li>
+        <li>
+          <!-- detached element / non-fully active elements -->
+          <p>
+            If <var>element</var> cannot navigate, return false.
```
MichaelKohler: We probably should drop this here, as it does not seem to apply to `<link>`: https://html.spec.whatwg.org/multipage/links.html#cannot-navigate. Given the list of triggers above, I don't think we need a "browsing-context connected" check either. Any thoughts on this?

specification/index.html
```diff
@@ -300,204 +300,187 @@ <h2>
...
+        <li>
+          <p>
+            If <var>element</var>'s [=Node/node document|node document=] is not
+            [=allowed to use=] the "monetization" feature, return false.
+          </p>
+        </li>
+        <li>
+          <p>
+            Let <var>context</var> be <var>element</var>'s [=Node/node
+            document|node document=]'s [=browsing context=].
```
MichaelKohler: We're not using context further on, so we can remove that as well.

specification/index.html
```diff
@@ -300,204 +300,187 @@ <h2>
         </p>
       </aside>
       <p>
-        The monetization keyword indicates a [=wallet address=] used to
-        monetize the document.
+        The <code>[=monetization=]</code> keyword indicates a [=wallet
```
sidvishnoi: This `monetization` refers to something else (the monetization concept in Model); not directly keyword. Let's remove this link.

sidvishnoi: Same for "monetized". Monetized should be a separate definition itself.

specification/index.html
```diff
@@ -300,204 +300,187 @@ <h2>
...
+        address=] used to monetize the document. It indicates that some aspect
+        of the document is [=monetized=].
       </p>
       <p>
-        The <code data-x="rel-monetization">monetization</code> keyword may be
-        used with <code>link</code> elements. This keyword creates an
-        <span data-x="external resource link">external resource link</span>.
+        The `monetization` keyword may be used with [^link^] elements. This
+        keyword creates an [=external resource link=]. This keyword is body-ok.
```
sidvishnoi:
```suggestion
        keyword creates an [=external resource link=]. This keyword is <a data-cite="html#body-ok">body-ok</a>.
```

specification/index.html
```diff
@@ -300,204 +300,187 @@ <h2>
...
       </p>
       <p>
-        The <code data-x="rel-monetization">monetization</code> keyword
-        indicates that some aspect of the document is monetized.
+        A user agent MUST NOT delay the load event for this link type.
```
sidvishnoi:
```suggestion
        A user agent MUST NOT delay the [=HTMLElement/load=] event for this link type.
```

specification/index.html
```diff
@@ -300,204 +300,187 @@ <h2>
...
+            [=external resource link=] that is already [=browsing-context
+            connected=], but was previously not obtained due to the
+            [^link/type^] attribute specifying an unsupported type, is set,
+            removed, or changed.
+          </p>
+        </li>
+      </ul>
+      <p>
+        The linked resource fetch setup steps for this type of linked resource,
+        given a [^link^] element <var>element</var> and [=request=]
```
sidvishnoi: Nit: `<var>element</var>` can be `|element|` in ReSpec. And we can give it type (on first instance like: `|element:HTMLLinkElement|`.

specification/index.html
```diff
@@ -300,204 +300,187 @@ <h2>
...
+        <var>request</var>, are:
+      </p>
+      <ol>
+        <li>
+          <p>
+            If <var>element</var>'s [^link/disabled^] attribute is set, stop
+            all previous [=payment sessions=] initiated by |element|'s
+            [=external resource link=], then return false.
```
sidvishnoi: I think this stopping of existing sessions doesn't relate to "fetch setup"?

sidvishnoi (commented):
For the `data-x` attributes:
I think the document was intentionally written such way so it can copy-pasted (monkey-patched) directly to HTML spec. Just FYI, we can translate it to HTML-spec syntax when time comes.

specification/index.html
MichaelKohler: We also link on other "monetized" occurrences, so I will keep it here for now. But I do agree with creating a separate definition for it in a follow-up.

specification/index.html
MichaelKohler: We need to have a place where we do it. Would be cleaner in the processing steps, but those steps are not run if the setup returns false. Then we'd end up with a session that keeps going even though it should not.

